PR #3, opened 2015-10-23
Honor consul.url and consul.addr from config file

jeinwag:
Looks like values for consul.addr and consul.url from the config file aren't passed to the consul client - fabio still tries to connect to localhost, so here's a quick fix.
Maybe these should rather be passed as arguments to NewWatcher (sorry, don't know much about go).


magiconair:
Yep. Good catch.

PR #4, opened 2015-10-23
Add Dockerfile

jeinwag:
Since fabio is something that I'd probably like to see running in a container, I've added a Dockerfile.
Yields a small image (about 16 MB).
Inspired by Jeff Lindsay's Dockerfile for registrator.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,13 @@
+FROM alpine:3.2
+ENTRYPOINT ["/bin/fabio"]
+
+COPY . /go/src/github.com/eBay/fabio
+
+RUN apk update
+RUN apk add -t build-deps go git \
+	&& cd /go/src/github.com/eBay/fabio \
+	&& export GOPATH=/go \
+	&& go get \
+	&& go build -o	/bin/fabio  \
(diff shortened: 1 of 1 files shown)


magiconair:
I'll try that myself first before I merge.

magiconair:
I've produced an official image from a scratch container also with the integrated ca-certificates.crt file to provide proper SSL support but I still need to test that with a valid cert. Could you try whether this image works for you since I'm not using Docker? If it is OK then I'll push my change.

    docker run -d -p 9999:9999 -p 9998:9998 -v $PWD/fabio.properties:/fabio.properties magiconair/fabio

jeinwag:
Thanks, looks goot to me!

magiconair:
I've moved the config file from `/fabio.properties` to `/etc/fabio/fabio.properties` so that you can store the configuration and one or more SSL certs in a single directory and just mount that instead of the individual files. I've updated the `magiconair/fabio:latest` and `magiconair/fabio:1.0.3` docker images accordingly. 

Docker support is now official and I'm closing this pull request (not merging)

PR #13, opened 2015-11-04
fix vet warning

juliendsv:
Just to avoid this warning when vetting:
go vet ./...
consul/watch_auto_config.go:69: missing argument for Printf("%v"): format reads arg 3, have only 2 args
exit status 1


magiconair:
Thx

PR #49, opened 2016-02-08
Fix up use of addr in service registration

r0p0s3c:
If the supplied addr has a host-ip part, use that, otherwise fallback to local ip detection


magiconair:
Yes, that makes more sense.

PR #58, opened 2016-02-19
Fix use of local ip in consul service registration

jeanblanchard:
When using the default registry.consul.register.ip, the service and health check ip were set to "<nil>",
making the health check fail.

This was regressed by PR #49, due to golang variable initialization magic.


magiconair:
Indeed. Thx

Issue #44, opened 2016-02-03
Strip prefix in the forwarded request


jeanblanchard:
Scenario: 
- I have fabio set up to listen to `http://fabioserver/`
- I have a route configured for a service, with prefix `/myservice`

When making a request to `http://fabioserver/myservice/index.html`, the service currently receives the forwarded request from fabio to `/myservice/index.html`
I would like to have the option to strip the prefix in fabio, so that the service would receive a request to `/index.html`

I currently have a homemade go proxy that does that, using `http.StripPrefix`. I may have the time to prepare a PR for this at some point, but no promises...

magiconair:
This is a duplicate of issue #2 for which I've decided not to support this. See the reasoning in the closing comment.

jeanblanchard:
OK I get the reasoning.

I would argue that this introduces a requirement that services need to know the route they serve.
You can't change the "mount" path for a service, say from `/myservice` to `/legacy/myservice` without changing the service itself (which may not be an option).

magiconair:
I would counter that argument that it is better to add another handler to the service than splitting up your configuration across services and load balancer. I know you can do and lots of other load balancers allow you to do it but in our own setup we have had enough additional complexity with this that this actually led to the development of fabio in the first place.

jeanblanchard:
... And I'm back...

It turns out I really need this after all. I tried to make all services have their prefix configurable, but this introduces a bit of complexity everywhere in the service, which I really want to avoid.
I think you'll agree that the services should not depend on the host where they are deployed. I believe the same is true for the path.
Stripping the prefix would allow services to work the same, regardless of the path where they are deployed.

Note that I do not suggest rewriting the urls in the responses (or maybe just Location headers), but instead encourage services to use relative urls.

Anyway, would you accept a pull request to add this option, or should I maintain my own fork ?

magiconair:
I agree on the host but not the path but that doesn't really matter. Please start this discussion with a suggestion on how this should work from the configuration perspective so that we can decide what the best approach is. No promises since I'd like to see how you want to solve that first and also want to think a bit about edge cases.

jeanblanchard:
On the configuration side, I'm thinking a global configuration flag in fabio.properties (`proxy.prefix.strip`?).

It's probably not necessary to allow each service/route to override the behavior. 

After a first look at the code, the global config case seems to be pretty straightforward. I'm not sure yet how to handle the websocket case, I'll look into it.

magiconair:
[...]

Issue #42, opened 2016-01-26
Feature: Allow weight tag in Consul


composer22:
"registrator" of paths into consul should also allow weight parameter to be passed so that traffic can be balanced to env needs. Some services may be running on nodes with less resources. Right now, rebalancing is distributed evenly.

Enhancement example:

```
docker run -d
-e "SERVICE_TAGS=urlprefix-/hello-world:0" 
--name hello_world_1  \
```

results in...

```
/hello-world
  /hello_world_1   100%
```

adding another specific weighted service...

```
docker run -d
-e "SERVICE_TAGS=urlprefix-/hello-world:20" 
--name hello_world_2  \
```

would result in...

```
/hello-world
  /hello_world_1   80%
  /hello_world_2   20%
```

magiconair:
I like the idea but would prefer a different syntax. Especially with the idea of the tcp proxy I'm thinking about using a URL for this:

`p-[scheme]://[host][:port][/path][?opt=val[&opt=val]...]`

Examples:

```
p-tcp://host:port
p-http://host:port/hello
p-https://host/hello
p-ws:///hello
p-/hello (equivalent to p-http:///hello)
p-/hello?weight=0.8
...
```

composer22:
@magiconair I like it. 

Let your imagination fly!

tiago-msilva:
Hi guys,

Any news on this matter?
I'm really interested in getting weights on TCP proxy mode.

Thanks!

nanoz:
Maybe we can help developing this feature ? If so, what would be the syntax ? `/path?weight=X` as a query paramater ?

magiconair:
@nanoz and @tiago-msilva I thought that this would be more work but the `route add` command already supports a `weight` option which does the right thing. The only thing missing was to expose this via the `urlprefix-` tag. You can now add a `weight=.x` option to specify a weight on a target, e.g. `urlprefix-:3306 weight=0.2 proto=tcp`

Sorry, @nanoz I did not want to take the opportunity for contributing away from you but was done pretty much when I looked at the the place which I wanted to point out to you. 

To illustrate this I've added a new `detail` option for the `log.routes.format` which you can use to dump the routing table in a different format.

```
./fabio -log.routes.format detail
2017/06/19 11:51:14 [INFO] Updated config to
+-- host=:3306
|   +-- path=
|       |-- addr=127.0.0.1:5001 weight 0.20 slots 2000/10000
[...]

PR #69, opened 2016-03-29
modifier: adding modifier tag concept

jonmorehouse:
When digging into fabio, I found the need to be able to route to upstreams based upon a "prefix" in a URI.

For instance, let's say I have a service I want to call `test` at `localhost:8000` and I want it accessible at `localhost:9999/test/` via fabio. I would like a request to `localhost:9999/test` to point to `localhost:8000/`, not `localhost:8000/test`.

I added the concept of a "modifier" tag, which allows you to trigger whether or not we strip the prefix when routing to a specific target. I'd love to hear any thoughts, its my first steps in this code base and I'm sure there are plenty of things to do better!

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -2,6 +2,7 @@ package proxy
 
 import (
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/eBay/fabio/config"
@@ -50,6 +51,16 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		// h = newWSProxy(t.URL)
 	default:
 		h = newHTTPProxy(t.URL, p.tr)
+	}
+
+	// strip the prefix, if the target was tagged properly
+	if t.StripPrefix {
+		pieces := strings.SplitN(r.URL.Path, "/", 3)
+		if len(pieces) == 3 {
+			r.URL.Path = pieces[2]
+		} else {
+			r.URL.Path = "/"
+		}
 	}
 
 	start := time.Now()
--- a/route/route.go
+++ b/route/route.go
@@ -3,6 +3,7 @@ package route
 import (
 	"fmt"
 	"net/url"
+	"regexp"
 	"sort"
 	"strings"
 
@@ -10,6 +11,8 @@ import (
 
 	gometrics "github.com/rcrowley/go-metrics"
 )
+
+var StripPrefixRegex *regexp.Regexp = regexp.MustCompile("modifier-/strip-prefix")
 
 // Route maps a path prefix to one or more target URLs.
 // routes can have a share value which describes the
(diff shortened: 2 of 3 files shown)


proxy/proxy.go
```diff
@@ -52,6 +53,16 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		h = newHTTPProxy(t.URL, p.tr)
 	}
 
+	// strip the prefix, if the target was tagged properly
```
jonmorehouse: I was digging through this, and it seems like the `Target` object could have a `Modify` method which accepts an `http.Request` object and modifies the request to its liking. In the case that we had more `modifier` tags, this could be quite useful

proxy/proxy.go
```diff
@@ -52,6 +53,16 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	if t.StripPrefix {
+		pieces := strings.SplitN(r.URL.Path, "/", 3)
+		if len(pieces) == 3 {
+			r.URL.Path = pieces[2]
```
jonmorehouse: This needs refactoring, was just getting something working :smile:

route/route.go
```diff
@@ -11,6 +12,8 @@ import (
 	gometrics "github.com/rcrowley/go-metrics"
 )
 
+var StripPrefixRegex *regexp.Regexp = regexp.MustCompile("modifier-/strip-prefix")
```
jonmorehouse: Would love to hear thoughts on naming here. I'm not overly opinionated, I also considered `urlmodifier-/strip-prefix` and `modifier-/strip`

route/route.go
```diff
@@ -51,7 +54,24 @@ func (r *Route) addTarget(service string, targetURL *url.URL, fixedWeight float6
...
+	stripPrefix := false
+	// by passing a tag of `url-prefix-host/modifier#strip-prefix you can
+	// drop whatever prefix the service is namespaced by
+	for _, tag := range tags {
+		if StripPrefixRegex.MatchString(tag) {
+			stripPrefix = true
+		}
+	}
+
+	t := &Target{
```
jonmorehouse: What do you think about a `NewTarget` method, which accepts the raw parameters and makes some decisions based upon which `tags` are specified?

magiconair:
Hi @jonmorehouse, this has been requested multiple times in #2 and #44 which is still open. Personally, I think stripping the URL prefix is the wrong approach since it moves essential configuration from your application into the load balancer and it makes your application behave differently with and without load balancer. IMO, in a lot of cases adding an additional handler isn't that much work but it makes the routing explicit. 

However, I get that some people would like to this support for this and there are some edge cases where adding handlers is not easy. I've outlined a different syntax for the `urlprefix` tag in #42 which would allow this option more organically. Feel free to work on that and build the strip prefix option on top of that.

leprechau:
This is absolutely needed in our environment.  We're currently using a very complex consul-template design that is based on tags including "proxy-standard" and "proxy-nostrip".  Our standard tag does a regrep on the request and strips the route prefix before passing it on the the backend whereas the nostrip does the opposite.

magiconair:
@jonmorehouse I'm going to close this PR as the strip prefix feature has been implemented for #44 and shipped in v1.3.7. In any case I want to thank you for your submission to show interest in the project and this feature.

PR #72, opened 2016-03-31
Proxy header handling 2

smancke:
Here is a more clean implementation of the former pull request #67.
Please look over it and tell me, what you think about it, now.

I was unsure, if I should add a user documentation on the headers to the readme?
Maybe it's to much detail for the readme?


proxy/http.go
```diff
@@ -8,6 +8,16 @@ import (
 
 func newHTTPProxy(t *url.URL, tr http.RoundTripper) http.Handler {
 	rp := httputil.NewSingleHostReverseProxy(t)
+
+	originalDirector := rp.Director
```
magiconair: Can you explain better in the comment why this is necessary and which problem it solves? Also, `s/originalDirector/d/`

proxy/http_test.go
```diff
@@ -0,0 +1,36 @@
...
+// with the target host as host header,
+// and not with the ip of the request to the proxy itself
+func TestCorrectHostHeader(t *testing.T) {
+
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Write([]byte(r.Host))
+	}))
+
+	defer server.Close()
+	serverUrl, _ := url.Parse(server.URL)
```
magiconair: s/serverUrl/serverURL/

proxy/http_test.go
```diff
@@ -0,0 +1,36 @@
...
+	}
+	proxy := newHTTPProxy(serverUrl, tr)
+
+	req, _ := http.NewRequest("GET", "http://example.com:666", nil)
+	res := httptest.NewRecorder()
+
+	proxy.ServeHTTP(res, req)
+
+	if serverUrl.Host != res.Body.String() {
+		t.Errorf("expected host was %v, but got %v", serverUrl.Host, res.Body.String())
```
magiconair: output pattern is `got X want Y`

proxy/http_test.go
```diff
@@ -0,0 +1,41 @@
...
+
+	req, err := http.NewRequest("GET", "http://example.com:666", nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	proxy.ServeHTTP(httptest.NewRecorder(), req)
+
+	if want := serverURL.Host; want != got {
+		t.Errorf("want host %q, but got %q", want, got)
```
magiconair: output pattern is `got X want Y`

proxy/proxy_integration_test.go
```diff
@@ -0,0 +1,43 @@
...
+		RequestURI: "/",
+		Header:     http.Header{"X-Forwarded-For": {"3.3.3.3"}},
+		RemoteAddr: "2.2.2.2:666",
+		URL:        &url.URL{},
+	}
+
+	proxy.ServeHTTP(httptest.NewRecorder(), req)
+
+	if want := "3.3.3.3, 2.2.2.2"; want != got {
+		t.Errorf("want %v, but got %v", want, got)
```
magiconair: `got X want Y`

proxy/proxy_integration_test.go
```diff
@@ -0,0 +1,43 @@
...
+	"net/url"
+	"testing"
+
+	"github.com/eBay/fabio/config"
+	"github.com/eBay/fabio/route"
+)
+
+func TestProxyProducesCorrectXffHeader(t *testing.T) {
+	got := "not called"
+	mockServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
magiconair: `s/mockServer/server/` to be consistent with your other test.

proxy/proxy_integration_test.go
```diff
@@ -0,0 +1,43 @@
...
+		got = r.Header.Get("X-Forwarded-For")
+	}))
+	defer mockServer.Close()
+
+	setRoutingTable(mockServer.URL)
```
magiconair: pls inline this

proxy/proxy_integration_test.go
```diff
@@ -0,0 +1,43 @@
...
+	tr := &http.Transport{Dial: (&net.Dialer{}).Dial}
```
magiconair: Either single line in both tests or multi-line in both tests

proxy/util_test.go
```diff
@@ -11,92 +10,109 @@ import (
...
+			http.Header{"X-Forwarded-Host": {"original.com"}},
+			"",
+		},
+		{"don't overwrite X-Forwarded-Proto header, if present",
+			&http.Request{RemoteAddr: "1.2.3.4:5555", Header: http.Header{"X-Forwarded-Proto": {"https"}}},
+			config.Proxy{},
+			http.Header{"X-Forwarded-Proto": {"https"}},
+			"",
+		},
+		{"set X-Real-Ip header, if not presen",
```
magiconair: typo: present

magiconair:
Looks OK. Some minor nitpicks in formatting and naming. Is there a way to squash the commits into one before submitting? I don't need/want the full history on how we got there. (used to gerrit)

smancke:
OK, I have done the style changes.
Squashing everything together to one commit is no problem, if everything is final.

But I got one question in mind, concerning the host header change, I made to the 
director of the SingleHostReverseProxy.
I think, my implementation is on the http standard and also the common way how e.g. nginx behaves. But, if someone is relying on the current behavior of fabio, than it's code may break. It would be an option, to make the behavior configurable with the current state as default. But this would mean that the default behavior of fabio would stay in being not correct.
So the decision is: Correct by default  vs.  full backwards compatible.

magiconair:
If I understand the host header change correctly then this behaves as follows:

Given the route `route add srv xxx.com/foo http://yyy.com/` the current implementation would forward with the `Host` header set to `xxx.com` and your change would change that to `yyy.com`, correct?

If that is the case than I cannot merge this since it would break our app big time. We rely on the original host header to be present. This discussion is along the same line I have with people asking for path prefix stripping. By doing this you move a central piece of configuration into the LB and you rely on this function to be there. Without that your application does not work. 

fabio is trying hard not to do that by not offering these options as the app should work the exactly same with and without fabio. This means that an app should handle all routes and it should also understand all host names. I understand that other LBs offer this but I think this should be solved in the app itself. Please let me know whether I understood this correctly.

magiconair:
Also, I don't think that this behavior is "correct" since it depends on what the LB is supposed to be doing. Sure, you can rewrite the host header but if the application needs to generate links that contain the full external host name then you would have to configure that separately since it can't extract that information from the request anymore. IMO, the whole path mangling is the wrong approach. Just like NAT on the IP level. It works but it causes all kinds of problems downstream.

smancke:
Hi @magiconair magiconair,
yes, you understood this correctly.

In my opinion it would be a bad practice, not to rewrite the host header, since this is how http usually woks in such an architecture. If an application wants to construct the url's, it has to respect the Forwading header. Every serious framework is doing so. Even without rewriting, you have to respect the headers, for the correct protocol scheme (http/https). 

But the world is full of opinions ;-)
So, please decide, how it should be:
a) Should I make the rewrite of the host header configurable, with the current behavior as the default? 
b) Or should I remove the host header rewriting from the pull request?

magiconair:
Lets leave it out for now. I need to think about this a bit more and TBH so far nobody has complained about it but I'd like to stick to RFCs where possible. Can you add another issue about this? Then we can have that discussion there. 

If I understand correctly I should be able to do the squashing in github now. Lets try that. 

Thank you

smancke:
Here #74 is a new pull request with the squashed commits, excluding those which are relevant to host header. (I also removed the code for X-Forwarded-Host and so, since it is only needed if we would rewrite it.)
 
An Issue for the Host Header thing is created at: #75

PR #81, opened 2016-04-14
Allow you to combine custom metrics prefix with default prefix.

md2k:
Very useful when you have single Graphite and split environments with namespaces.

`metrics.prefix = myNameSpace.default`
  Example:
    '`dev.default`' will be converted to prefix as follow '`dev.<hostname>.<executable>`'
    '`qa.default`' will be converted to prefix as follow '`qa.<hostname>.<executable>`'
    '`prod.default`' will be converted to prefix as follow '`prod.<hostname>.<executable>`'

### DIFF
--- a/fabio.properties
+++ b/fabio.properties
@@ -242,6 +242,11 @@
 #
 # When set to "default" the prefix is <hostname>.<executable>
 #
+# You hape option to combine custom prefix with default if metrics using  namespaces
+# Example: 
+#   'dev.default' will be converted to prefix as follow 'dev.<hostname>.<executable>'
+#
+#
 # The default is
 #
 # metrics.prefix = default
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -1,14 +1,17 @@
 package metrics
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"log"
 	"net"
 	"net/url"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
+	"text/template"
 	"time"
 
 	"github.com/cyberdelia/go-metrics-graphite"
@@ -17,6 +20,12 @@ import (
 )
 
 var pfx string
+
+type Prefix struct {
+	Fqdn     string
+	Hostname string
+	Exe      string
+}
 
 // ServiceRegistry contains a separate metrics registry for
 // the timers for all targets to avoid conflicts
@@ -34,8 +43,20 @@ func Init(cfgs []config.Metrics) error {
 
 func initMetrics(cfg config.Metrics) error {
 	pfx = cfg.Prefix
+	prefix := initPrefixTemplate()
 	if strings.Contains(pfx, "default") {
 		pfx = strings.Replace(pfx, "default", defaultPrefix(), 1)
+	} else {
+
+		t := template.New("Prefix template")
+		t, err := t.Parse(pfx)
+		if err != nil {
+			fmt.Println("Fatal error ", err.Error())
+			os.Exit(1)
+		}
+		var doc bytes.Buffer
+		t.Execute(&doc, prefix)
+		pfx = doc.String()
 	}
 
 	switch cfg.Target {
@@ -85,6 +106,31 @@ func defaultPrefix() string {
 	}
 	exe := filepath.Base(os.Args[0])
 	return clean(host) + "." + clean(exe)
+}
+
+func initPrefixTemplate() *Prefix {
+	// Init Template variables
+	var fqdn string
+	hostname, err := hostname()
+	if err != nil {
+		log.Fatal("[FATAL] ", err)
+	}
+	out, err := exec.Command("hostname", "-f").Output()
+	if err != nil {
+		log.Printf("[WARN] Couldn't determine hostname fqdn, failing back to os.Hostname()", err)
+		fqdn = hostname
+	} else {
+		fqdn = strings.Trim(string(out), "\n\r\t")
+	}
+	exe := filepath.Base(os.Args[0])
+
+	// Assign template variables to Prefix structure
+	prefix := Prefix{
+		Fqdn:     clean(fqdn),
+		Hostname: clean(hostname),
+		Exe:      clean(exe),
+	}
+	return &prefix
 }
 
 func initStdout(interval time.Duration) error {
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -34,8 +34,8 @@ func Init(cfgs []config.Metrics) error {
 
 func initMetrics(cfg config.Metrics) error {
 	pfx = cfg.Prefix
-	if pfx == "default" {
-		pfx = defaultPrefix()
+	if strings.Contains(pfx, "default") {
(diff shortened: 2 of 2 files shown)


magiconair:
This will break with `defaultEnv.default`. The whole magic term `default` is broken. It might make more sense to make the default a set of variables which are pre-defined. Then the default just becomes a format string, e.g. `metrics.prefix = ${hostname}.${exe}` I have to think about that.

md2k:
Hi Frank, you right and templates for prefixes will be more elegant way, but solution above was faster and allow me solve the problem with namespaces instantly with minimal changes of Fabio code and reduce amount of possible problems with it.
I'll also check other possibilities to do it better way, since i'm in demands of that feature.

md2k:
Also need add possibility to use short hostname and fqdn (hostname -f), because currently it use only short hostname of the server.

magiconair:
The simplest workaround is to set an ENV variable `FABIO_PREFIX` and change `metrics.prefix = ${FABIO_PREFIX}`. That should work out of the box.

md2k:
@magiconair check my latest commit please. templates use native text/template library
and now custom prefix can be set next way: `metrics.prefix = myapp.{{.Fqdn}}.{{.Exe}}`
Not final variant, but feedback will be useful.
Your advice to use ENV, not very suitable for me at this moment.

magiconair:
Hmm, using `text/template` is an interesting approach. The code needs some cleanup but the biggest concern is that `hostname -f` won't work on Solaris the way you think it does - and there are people running fabio on Solaris. With this we could also make the `clean` function available through the template.

fabio.properties
```diff
@@ -242,6 +242,11 @@
 #
 # When set to "default" the prefix is <hostname>.<executable>
 #
+# You hape option to combine custom prefix with default if metrics using  namespaces
```
magiconair: The comment does not match the new behavior using `text/template`

metrics/metrics.go
```diff
@@ -18,6 +21,12 @@ import (
 
 var pfx string
 
+type Prefix struct {
+	Fqdn     string
```
magiconair: s/Fqdn/FQDN/

metrics/metrics.go
```diff
@@ -34,8 +43,19 @@ func Init(cfgs []config.Metrics) error {
 
 func initMetrics(cfg config.Metrics) error {
 	pfx = cfg.Prefix
+	prefix := initPrefixTemplate()
```
magiconair: You should move the whole prefix init code to a helper function

metrics/metrics.go
```diff
@@ -34,8 +43,19 @@ func Init(cfgs []config.Metrics) error {
...
 	if pfx == "default" {
 		pfx = defaultPrefix()
+	} else {
+		t := template.New("Prefix template")
+		t, err := t.Parse(pfx)
+		if err != nil {
+			fmt.Println("Fatal error ", err.Error())
```
magiconair: This should be `log.Fatalf("[FATAL] prefix template failed: %s", err)

metrics/metrics.go
```diff
@@ -87,6 +107,31 @@ func defaultPrefix() string {
...
 }
 
+func initPrefixTemplate() *Prefix {
+	// Init Template variables
+	var fqdn string
+	hostname, err := hostname()
+	if err != nil {
+		log.Fatal("[FATAL] ", err)
+	}
+	out, err := exec.Command("hostname", "-f").Output()
```
magiconair: This will set the hostname to `-f` on Solaris

md2k:
About `hostname -f` on solaris i know, this why it have error handler where it use normal hostname in case of error, but not have solaris to test this part, but we can add OS detection for example for Win/Sol to fallback FQDN to hostname. i'll check more precisely and cleanups/fixes.

favoretti:
@md2k calling system binaries in portable software is a huge no-no. Something in line of `net.LookupHost()` on your non-localhost IP would probably do the trick.

md2k:
@favoretti , i know, but in GoLang i not found any other way to do so. if you have other ideas , shoot :)

magiconair:
@md2k Can you explain why using command line flags or environment variables is not an option for you?

md2k:
@magiconair because then it require take care about recipes to manage dynamic variables and use those variables in application declaration files. I think this killing some part of automation, because then i should ensure that those variables are set. If variables static like `environment=production` or `GOGC` iti s fine, but if variables unknown before application start, like FQDN of instance, or IP where application listening i should take care with some amount of scripting around or keep in my deployment tool map of variable=value per host, definitely i wanna escape those scenarios if possible and this why i came with idea which allow use templates in prefix configuration. All depends from internal structure of projects, in my case i trying to minimize amount of scripting around application during its deployment and startup.
But i agreed with @favoretti, and i not should use binary exec because it is OS dependent.

magiconair:
@md2k If you can run `GOGC=400 /usr/local/bin/fabio` then I think you can also run `GOGC=400 METRIC_PREFIX=${ENV}.$(hostname -f) /usr/local/bin/fabio` the main problem with this approach is that you probably want to clean the hostname separately from the env so this won't do what you expect. I like the template approach in general but the current implementation will not work since it isn't portable. I have to think about this a bit more.

favoretti:
One could do something like:

```
func externalIP() (string, error) {
	addrs, err := net.InterfaceAddrs()
	if err != nil {
		return "", err
	}
	for _, addr := range addrs {
		var ip net.IP
		switch v := addr.(type) {
		case *net.IPNet:
			ip = v.IP
		case *net.IPAddr:
			ip = v.IP
		}
		if ip == nil || ip.IsLoopback() {
			continue
		}
		ip = ip.To4()
		if ip == nil {
			continue // not an ipv4 address
		}
		return ip.String(), nil
	}
	return "", errors.New("are you connected to the network?")
}
```

And then basically convert that IP to a hostname. Not 100% fool-proof, but better than calling `hostname`.

md2k:
`METRIC_PREFIX=${ENV}.$(hostname -f)` will work good only for bash based bootstrap applications like daemontool or similar with ability to accept and execute bash commands, we for example use another approach to start applications where i unable to pass bash executable commands, unfortunately. 
I agreed that my solution not portable because i use `hostname -f`, and i think we can skip that part for now and just cleanup template solution. If you fine with `text/template` functionality i can try to clean it or find another template solution.
@favoretti, your solution may work but then you should ensure about IP address passed to function and that your PTR record correctly maintained. I think lets skip hostname detection. more easy change instances hostnames to fqdn and simply allow to user customise `prefix` by template.

magiconair:
Maybe we want this to default to the value of `proxy.localip`

md2k:
if you bind `proxy.addr` to `0.0.0.0` then you should care about `proxy.localip` manually or no?

magiconair:
`proxy.localip` is set automatically to the first address it finds.

md2k:
may cause problems if instance have multiple interfaces with more then one network and if IPs have different PTR records or /etc/hosts records
`1.1.1.1 server.external.example.com`
`2.2.2.2 server.internal.example.com`
it means that `localip` can be set differently between starts and have different hostnames in metrics, chances not so big, but exists.

PR #93, opened 2016-05-02
Add glob path matching (an alternative to default prefix matching)

dkong:
We have a use case that requires being able to register a backend service with a glob style pattern `/foo.*` and have it handle all routes such as `/foo.bar`, `/foo.bar.baz`, etc.

This PR retains the original prefix matching behavior but adds an option to use glob matching.  This also opens up the possibility to add regex matching down the line if it's useful.

Thanks.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -29,6 +29,7 @@ type UI struct {
 
 type Proxy struct {
 	Strategy              string
+	Matcher               string
 	MaxConn               int
 	ShutdownWait          time.Duration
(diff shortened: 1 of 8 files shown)


magiconair:
Interesting idea. This might solve a couple of other issues for people as well. Code looks clean on first glance and I like the integration. I'll spend a bit more time reviewing this but this looks like it is going in.

Thanks for this.

magiconair:
Looks good. I'll update the docs.

Issue #80, opened 2016-04-11
access logging


rutchkiwi:
Hi! 

First off, thank you for making this, it's a really cool project!

My question is quite simple, does fabio support access logging? nginx for instance can print all requests passing through and their responses to stdout or a file. Would be quite useful to me, since I'd then would not need to implement this in all downstream services.

magiconair:
Not yet but certainly doable. Just log to local disk?

rutchkiwi:
thx for the quick reply!
In my case I'd like to log to stdout, since I'm running the docker image and all my other containers send their logs off to a log aggregation service using a docker log driver.

magiconair:
so access log to stdout, normal log to stderr?

rutchkiwi:
That would be perfect for me

magiconair:
Any preference on the format?

rutchkiwi:
No strong preference really, but json would be easy to parse. The data I capture in other places would be 
- the http method/hostname/path
- the response time & status
- which (upstream) server it got routed to
- the local time
- the user agent

Of course for a first version it would be really nice just having the first two! Let me know if there is anything I could do to help with this. :)

osigida:
I guess, the format should be configurable, say we'd like to have some headers as well.

Here is implementation from logback, might be useful: http://logback.qos.ch/access.html

sfleiter:
What is still missing to be able to merge this?
Looking forward to the feature.

magiconair:
There was quite a bit of work to do AFAIR. I'm on vacation for another week and should have time to pick some of the slack early December. 

Frank Schröder

sfleiter:
Thanks a lot for your fast feedback Frank,
hope you have a wonderful vacation!

sielaq:
+1 for access logs, also to a file too

sielaq:
![please](http://danhairfield.files.wordpress.com/2011/12/2005_07_21_shrek_cat1.jpg)

ElliotG:
Also wanted to add that access logging would be super awesome.

leprechau:
I'm not sure I see the rationale to have this goto a file natively.  Running fabio under systemd, upstart or in a docker container allows you to easily get any stdout/stderr into a file with standard tools.

magiconair:
Lets first get this working. Writing to a file is trivial but I agree with @leprechau that this shouldn't be necessary in this day and age.

magiconair:
[...]

PR #131, opened 2016-07-24
Feature/support access logging

beyondblog:
Add access logging like #80 

Written to disk, format is configurable like nginx log_format

E.g
```
fabio.properties

proxy.log.enable = true
proxy.log.file.path = access.log
proxy.log.format = $remote_addr $time $request $body_bytes_sent $http_referer $http_user_agent $server_name $upstream_addr $upstream_response_time $request_args
```

support format syntax:

```
$remote_addr      
$time 
$request
$body_bytes_sent
$http_referer 
$http_user_agent 
$server_name 
$upstream_addr 
$upstream_response_time 
$request_args
```

But the response state does not support, because I do not know how to get.


magiconair:
Thanks. I'll have a look.

magiconair:
@beyondblog I had a look and here are my comments:

I like the approach of creating a list of functions which create the log line depending on the format. The implementation is abusing the types a bit and makes too many allocations by using string concatenation instead of a `bytes.Buffer` from a `sync.Pool`

I'd replace the types with something like this:

```
type value func(w io.Writer, t time.Time, r *http.Request)

var values = map[string]value{
	"$remote_addr": func(w io.Writer, t time.Time, r *http.Request) {
		w.WriteString(r.RemoteAddr[:strings.Index(r.RemoteAddr, ":")])
	},
	"$time": func(w io.Writer, t time.Time, r *http.Request) {
		w.WriteString(t.Format(time.RFC3339))
	},
	"$request": func(w io.Writer, t time.Time, r *http.Request) {
		w.WriteRune('\"')
		w.WriteString(r.Method)
		w.WriteRune(' ')
		w.WriteString(r.URL.Path)
		w.WriteRune(' ')
		w.Write(r.Proto)
		w.WriteRune('\"')
	},
}

var pattern []value
```

In addition, I'd use a `sync.Pool` to re-use the buffers since this code sits in the hot-path. 

```
var BufSize = 1024

var pool = sync.Pool {
	New:func() interface{} {
		return bytes.NewBuffer(make([]byte, 0, BufSize))
	},
}

func logger(w io.Writer, format string, t time.Time, r *http.Request) {
	b := pool.Get().(*bytes.Buffer)
	b.Reset()
	for _, p := range pattern {
		p(b, t, r)
		b.WriteRune(' ')
	}
	b.Truncate(b.Len()-1) // drop last space
	_, err := w.Write(b.Bytes())
}
```

I'd also drop the lazy pattern parsing and prefer an `Init()` function which parses the pattern and returns an error. This way you save a `nil` check on every request.

Then, I would not use a `log.Logger` for the output since you are already creating the full log line including time stamp. Just pass in an `io.Writer`.

I am also not sure if fabio should write the access log to a file. The normal log is written to `stderr` which means you need to capture and redirect it anyway. `fabio` should also be run under a supervisor which restarts it (daemontools, systemd, supervisor, ...) so that you could redirect the output to a file and let someone else handle log rotation. What if you want to ship the access logs via rsyslog over the network? What if fabio runs in a docker container. For now I would just send them to `stdout` which would save another allocation since you can then write the content of the buffer directly to stdout instead of converting it to a string first.

`proxy.log.enable` is redundant. `proxy.log.file.path` should be named `proxy.log.target` and should default to empty which disables access logging. The other option should be `stdout`. 

I might also drop the `$` from the format variables since it doesn't serve any purpose but it might increase readability. 

I might want to have a look at the speed of `time.Format(time.RFC3339)` and check if a special implementation would be better. Again, keep in mind that we're driving fabio with a couple of thousand requests per second and adding lots of allocations to every request will just put more pressure on the GC.

beyondblog:
@magiconair Thanks.Your suggestion is very good.I will try to correct.

magiconair:
Looks better. I've added some comments. Please drop the `enabled` flag since it is redundant. If you don't want to log you should not have a logger in the first place. I'd also move the logger creation into the `main` function since it can `FATAL` and then inject it into the Proxy.

magiconair:
@beyondblog I think a logger package is a good idea. Maybe like this?

```
package logger

var BufSize = 1024

type value func(b *bytes.Buffer, t time.Time, r *http.Request)

var values = map[string]value { ... }
 
type Logger struct {
    // p stores the log pattern as a list of value functions
    p []value

    // mu guards w
    mu sync.Mutex

    // w is the log destination
    w io.Writer
}

func (l *Logger) New(w io.Writer, format string) (*Logger, error) {
    if w == nil { return nil, nil }
    p, err := parse(format)
    if err != nil { return nil, err }
    return &Logger{p, w}
}

func parse(format string) ([]value, error) {
    var vv []value
    for _, f := range strings.Fields(format) {
        v := values[f]
        if v == nil { return nil, fmt.Errorf("invalid log field %s", f) }
        vv = append(vv, v)
    }
    return vv, nil
}

func (l *Logger) Log(t time.Time, r *http.Request) {
    b := pool.Get()
    b.Reset()
    for _, v := range l.p {
        v(b, t, r)
        b.WriteRune(' ')
    }
    b.Truncate(b.Len()-1) // drop last space
    b.WriteRune('\n')
    l.mu.Lock()
    l.w.Write(b.Bytes())
    l.mu.Unlock()
    pool.Put(b)
}
```

Could you also add some table driven tests for the value functions and the parsing? I usually write them like this:

```
func TestLog(t *testing.T) {
	ts := time.Date(2016, 1, 1, 0, 0, 0, 0, time.UTC)
	tests := []struct {
		req    *http.Request
		format string
		out    string
	}{
		{},
		{&http.Request{Host: "abc.com"}, "req_host", "abc.com"},
                ...
	}

	for i, tt := range tests {
                b := new(bytes.Buffer)
		l, err := logger.New(b, tt.format)
                if err != nil { t.Fatalf("%d: got %v want nil", i, err) }
                l.Log(ts, tt.req)
		if got, want := string(b.Bytes()), tt.out; got != want {
			t.Errorf("%d: got %q want %q", i, got, want)
		}
	}
}
```

magiconair:
Apologies to the watchers for spamming you with code-review comments. I'm used to the gerrit workflow where all comments get published in one go.

beyondblog:
@magiconair Thanks. I will improve my code under your suggestion.

magiconair:
@beyondblog I've taken your PR and refactored it a bit since the logger is sitting in the hot path and I'd prefer it not to allocate memory since that would put pressure on the GC. 

Unfortunately, I still haven't figured out how to collaborate on PRs the same way as it is possible to do this on gerrit. Therefore, I've created a new branch, squashed your commits and rebased them on `master` and added my cleanups on top of them. If you know how to do this better on GH then pls let me know.

Here is what I did:

* Add a benchmark and added some comparisons
* Use `bytes.Buffer` instead of `io.Writer` so that I can use the `WriteXXX` methods directly
* Inlined all calls to `fmt.Sprintf()` and `time.Format()` to avoid allocations. For this I've written a custom `atoi` function which does not alloc. 
* I've added more tests to cover all log fields and I am not sure if the way this is currently implemented yields to the correct results. However, I was focussing mainly on having the tests and the benchmarks and not on the correctness of the tests.

```
# before
BenchmarkLog-8	 1000000	      1807 ns/op	      80 B/op	       6 allocs/op

#after
BenchmarkLog-8	 2000000	       919 ns/op	       0 B/op	       0 allocs/op

$ benchcmp 0-baseline.txt 3.5-inline-time-format.txt
benchmark          old ns/op     new ns/op     delta
BenchmarkLog-8     1807          919           -49.14%

benchmark          old allocs     new allocs     delta
BenchmarkLog-8     6              0              -100.00%

benchmark          old bytes     new bytes     delta
BenchmarkLog-8     80            0             -100.00%
```

beyondblog:
@magiconair  Thanks. I look forward to this feature.

roman-gliffy:
Would be nice to have logs.
@magiconair, any suggestions how to deal with the absent of logs in production for the time being?

magiconair:
Yeah, it's still on my plate but I got sidetracked. Without Fabio doing the logging you've got only two options. Either your apps do the logging or you put another proxy in front of Fabio. However, this won't provide you with the instance it got routed to.

rokka-n:
Yeah, right.  Having backends info is what logging would be very useful for. However, I was looking at the docs and there is trace mode. Which kinda provides this information.
 
Is it possible to enable it, maybe dynamically for all requests?

magiconair:
@roman-gliffy FYI, #80 is tracking this and I've got a patch ready for testing. Should be merged soon.
@rokka-n The patch also supports logging the upstream server.

rokka-n:
👍 Thank you!

PR #143, opened 2016-08-16
Update parse.go

juliangamble:
The error text when you don't have a trailing slash isn't clear. This is to make it super-obvious!


magiconair:
:)

Issue #147, opened 2016-08-25
Support multiple metrics libraries


postwait:
Current the metrics subsystem doesn't support histogram reporting of request latencies.  Can we make the metrics reporting interface such that the implementation can be changed out for "gometrics"-style recording that supports histograms?

I'd like to integrate circonus-gometrics, but layered above the go-metrics currently used, all good measurement data is destroyed.

maier:
just discussed a shim approach which sounds like it will solve the issue. if there is a metric interface in fabio itself then swapping out backends which want/support full raw measurement values will be easier.

magiconair:
@postwait and @maier I've moved the metrics library implementation behind an interface in the `metrics/metricslib` package where I've extracted the methods from the go-metrics library I'm currently using. This should make it possible to plug in any metrics library you want as long as it supports this interface. If the interface is not sufficient for what you are trying to do feel free to enhance it but keep it simple and generic if possible. Looking forward to the PRs :)

magiconair:
I think I missed a spot. The current implementation distinguishes between two different namespaces for metrics: the service metrics and the general fabio metrics (like `requests`). Service metrics get purged when the service or an instance disappears. This was implemented by using two different registries but that abstraction got lost with this change. I need to add that back in or make the purge code aware of what is a service metric and what isn't. Just keep that in mind when implementing your change.

maier:
Cool, thanks, I'm working off that branch to do an integration.

magiconair:
@maier My previous approach was too complex. The whole notion of the `metrics` package providing two registries was somewhat broken which made the testing more complex than needed to be. I've dropped the `metricslib` and the `gometrics` package and rolled this all into the `metrics` package. There you can now create multiple registries each with a single name space. The `main.go` then sets up the two name spaces for global and service metrics. That makes it cleaner and simpler IMO but you need to change something in your code. Sorry about that. 

Supporting multiple different providers could then be implemented by adding a `MultiRegistry` analogous to the `io.MultiWriter`

maier:
I'll pull down the latest commit and check it out to see what the delta is between what I've got working and the updated method.

magiconair:
Merged to master.

magiconair:
@maier and @postwait please file follow-up tickets for issues with the circonus implementation. This is now on master. It would be great if you could test it and let me know whether it works.

magiconair:
@maier Thanks for your help with this.

maier:
[...]

PR #149, opened 2016-08-30
Issue #147: Support multiple metrics libraries

maier:
Add Circonus


magiconair:
What happens when a service or an instance of a service appears and then disappears again? Will the circonus library stop reporting metrics for this instance? That's the reason behind the `Names()` and `Unregister()` functions since they allow fabio to trim the reported metrics to the current set of services and instances.

maier:
Circonus doesn't keep a persistent set of metrics in memory (like gometrics does). The metrics are flushed after each submission. If a given metrics stops being used, it would no longer be sent to Circonus.

magiconair:
OK. As for the vendored libraries: Updating `vendor.json` is not enough. You need to run `govendor add` for all libraries and check the source of the libraries in. I usually do this in a separate commit labeled 'Vendoring github.com/... plus dependencies' and then one additional commit with all the squashed changes for the issue. I can squash commits in github now but then I'd squash your changes and the vendored libs together. Could you please create two commits with first the vendored libs and then your changes? `git push -f` is fine.

magiconair:
Check the last three commits on master for an example.

maier:
ok. so what you want is:

commit vendor.json and all vendor source updates as "Vendoring github.com/circonus-labs/circonus-gometrics plus dependencies"

commit with just my changes to source tree

pull request

Issue #125, opened 2016-07-11
Extended metrics


robholland:
We're interested in measuring the impact of switching a system over to fabio versus a haproxy setup, it would be very useful to have a Counter() per HTTP status code so we can check for behaviour versus our current setup, ensuring the 200/202/404 rates are equivalent.

Also it would be useful to have a Counter() for route-not-found responses (proxy.noroutestatus) as they won't be associated with a route target.

magiconair:
Sounds good. I'll see what I can do.

magiconair:
@robholland The patch I've uploaded should do the trick. 

```
    * Add 'noroute' counter for requests to non-existing routes
    * Add 'http.status.xxx' timer for responses with a specific status code
    * Add 'ws.conn' counter for counting number of open websockets
```

I'm a bit torn by keeping track of all status codes since this code is in the hot path. The generic implementation has to create the metric name and lookup the timer for every request which means 1-2 allocs and accessing a (global?) lock. 

A possible optimization would be to limit the number of status codes fabio keeps track of and make that configurable. Then I can pre-alloc the timers and don't have to look them up. 

If it is easy for you to run a load test to see whether that makes a difference then that would help a lot. I'll see what I can do on our end. For now, I've opted for the more useful implementation but I'll merge this after I've seen some numbers.

robholland:
Frank, we're going to have to halt our fabio experiments as our priorities have changed. If you're keen on getting this merged and need some testing then I'll try and make some time for it in a few weeks.

Thanks for your efforts getting us this far, hopefully we'll be back :)

magiconair:
@robholland No problem. I think this one has enough value. I'll run this in our environment first to see how it behaves. Let me know if/when I can help again.

magiconair:
Rebased this on top of #147

maier:
```go
type cgmCounter struct {
	metrics *cgm.CirconusMetrics
	name    string
}

func (m *cgmRegistry) GetCounter(name string) Counter {
	metricName := fmt.Sprintf("%s`%s", m.prefix, name)
	return &cgmCounter{m.metrics, metricName}
}

func (c cgmCounter) Inc(n int64) {
	c.metrics.IncrementByValue(c.name, uint64(n))
}
```

magiconair:
Merged to master.

PR #150, opened 2016-08-30
Issue #151: Add support for Circonus metrics

maier:
Adding support for sending metrics to Circonus.


fabio.properties
```diff
@@ -505,6 +506,46 @@
 # metrics.statsd.addr =
 
 
+# metrics.circonus.apiapp configures the API token app to use when
+# submitting metrics to Circonus. See: https://login.circonus.com/user/tokens
+# This is optional when ${metrics.target} is set to "circonus".
+#
+# The default is
+#
+# metrics.circonus.apiapp = "fabio"
```
magiconair: Don't use double quotes unless you want to send them. Also, if this is the default then you need to set this in `config/default.go`. Same for `metrics.circonus.apiurl`

metrics/circonus.go
```diff
@@ -0,0 +1,132 @@
...
+	"fmt"
+	"log"
+	"os"
+	"sync"
+	"time"
+
+	cgm "github.com/circonus-labs/circonus-gometrics"
+)
+
+type cgmProvider struct {
```
magiconair: s/cgmProvider/cgmRegistry/

metrics/circonus.go
```diff
@@ -0,0 +1,132 @@
...
+	metrics *cgm.CirconusMetrics
+	name    string
+}
+
+var (
+	circonus *cgmProvider
+	once     sync.Once
+)
+
+const serviceName = "fabio"
```
magiconair: Isn't this the value from `metrics.circonus.apiapp`? If not, then either make it configurable or inline it below instead of using `fmt.Sprintf`

metrics/circonus.go
```diff
@@ -0,0 +1,132 @@
...
+		cfg.CheckManager.Check.ID = circCheckID
+		cfg.CheckManager.Broker.ID = circBrokerID
+		cfg.Interval = fmt.Sprintf("%.0fs", interval.Seconds())
+		cfg.CheckManager.Check.InstanceID = host
+		cfg.CheckManager.Check.DisplayName = fmt.Sprintf("%s /%s", host, serviceName)
+		cfg.CheckManager.Check.SearchTag = fmt.Sprintf("service:%s", serviceName)
+
+		metrics, err := cgm.NewCirconusMetrics(cfg)
+		if err != nil {
+			initError = fmt.Errorf("metrics: unable to initialize Circonus %s", err)
```
magiconair: Don't you need to return here or can you start the metrics even though you have an error? This would be different than for the other metrics libs.

metrics/circonus.go
```diff
@@ -0,0 +1,132 @@
...
+		metrics.Start()
+
+		log.Print("[INFO] Sending metrics to Circonus")
+
+	})
+
+	return circonus, initError
+}
+
+// Names returns the list of registered metrics acquired
```
magiconair: Pls drop the interface comments. Just leave the comments that explain that circonus does not support this.

metrics/circonus_test.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	t.Log("\tRate1()")
+	rate := timer.Rate1()
+	if rate != 0 {
+		t.Errorf("Expected 0 got '%+v'", rate)
+	}
+}
+
+func TestAll(t *testing.T) {
```
magiconair: What exactly is this testing? How do you know that this test has passed?

metrics/circonus.go
```diff
@@ -0,0 +1,132 @@
...
+
+var (
+	circonus *cgmProvider
+	once     sync.Once
+)
+
+const serviceName = "fabio"
+
+// circonusBackend returns a provider that reports to Circonus.
+func circonusBackend(prefix string,
```
magiconair: s/circonusBackend/circonusRegistry/

metrics/circonus.go
maier: ok, will change.

fabio.properties
maier: Added a default for CirconusAPIApp to defaults. I reworded the description of the API URL parameter, I'd prefer to keep that default localized in the library so that if it changes there isn't a waterfall of PRs to update.

magiconair:
Could you also update the comment for the 2nd commit to `Issue #147: Add support to send metrics to Circonus`. I think if you force push the branch to your repo `git push -f` then this won't create additional commits in the PR. GH should just replace the current PR.

metrics/circonus.go
maier: doh! yeah, good catch, thanks.

fabio.properties
```diff
@@ -512,16 +512,18 @@
 #
 # The default is
 #
-# metrics.circonus.apiapp = "fabio"
+# metrics.circonus.apiapp =
```
magiconair: Hmm, if you set a default in `config/default.go` then this needs to be reflected here. Why did you remove it?

should be `metrics.circonus.apiapp = fabio`

magiconair:
I thinks this looks ok enough. I'll merge this manually since the `config/load_test.go` needs updating and there are one or two typos. Thanks for your patience with the code review.

magiconair:
@maier you seem to have forgotten to add `metrics.circonus.apikey` to the `fabio.properties`. Could you provide me with the snippet and then I'll add it. Thx

magiconair:
@maier or is this ok?

```
# metrics.circonus.apikey configures the API key to use when
# submitting metrics to Circonus. See: https://login.circonus.com/user/tokens
# This is optional when ${metrics.target} is set to "circonus".
#
# The default is
#
# metrics.circonus.apikey =
```

maier:
```
# metrics.circonus.apikey configures the API token key to use when
# submitting metrics to Circonus. See: https://login.circonus.com/user/tokens
# This is required when ${metrics.target} is set to "circonus".
#
# The default is
#
# metrics.circonus.apikey =
```

maier:
that's the one setting which is required, the rest are optional.

magiconair:
Got it. I'm still curious what `TestAll` is testing and how you determine that the test has passed. Could you explain, please?

maier:
testAll actually *does* the job, it connects to the api and performs all of the steps. verification is two-fold, if the code fails and then checking manually in the UI for the metric coming through. the code will fail in finding/creating the check, then i go over to the ui and verify that a) the fake metric was created and b) a submission came through for it.

magiconair:
ok, that's what I thought. This is an integration test which requires manual inspection. I'll leave it in for now but need to think about whether I would want to mock the API or check this another way. I don't  want to test the driver. 

As a side note: I would also like to merge the change from #125 which adds a counter to the metrics registry. 

```
 // Counter defines a metric for counting events.
 type Counter interface {
     // Inc increases the counter value by 'n'.
     Inc(n int64)
 }
```

How would I add this to the circonus registry?

maier:
```go
type cgmCounter struct {
	metrics *cgm.CirconusMetrics
	name    string
}

func (m *cgmRegistry) GetCounter(name string) Counter {
	metricName := fmt.Sprintf("%s`%s", m.prefix, name)
	return &cgmCounter{m.metrics, metricName}
}

func (c cgmCounter) Inc(n int64) {
	c.metrics.IncrementByValue(c.name, uint64(n))
}
```

magiconair:
OK, I just noticed that the registry methods return pointers (`&cgmTimer`) but the methods on the structs have value receivers. How does this even work? I guess in the current implementation it doesn't make a difference whether it is either/or but I think it should be consistent. I'll change the timer/counter methods to have pointer receivers as well (e.g. `func (t *cgmTimer) UpdateSince(...)`) unless you know a good reason why I shouldn't.

magiconair:
Did a manual merge of this PR to master.

PR #160, opened 2016-09-14
Metrics Prefix with templates

md2k:
This is actually same as https://github.com/eBay/fabio/pull/81
By mistake wiped out my brunch.
I removed modifications about external `hostname` execution, because it bit tricky to achieve same functionality across platforms, so i just moved `metrics.prefix` to templates same  way as  `metrics.names`.

```
# metrics.prefix configures the template for the prefix of all reported metrics.
#
# Each metric has a unique name which is hard-coded to
#
#    prefix.service.host.path.target-addr
#
# The value is expanded by the text/template package and provides
# the following variables:
#
#  - Hostname:  the Hostname of the server
#  - Exec:      the executable name of application
#
# The following additional functions are defined:
#
#  - clean:     lowercase value and replace '.' and ':' with '_'
#
# Template may include regular string parts to customize final prefix
#
# Example:
#
#  Server hostname: test-001.something.com
#  Binary executable name: fabio
#
#  The template variables are:
#
#  .Hostname =  test-001.something.com
#  .Exec = fabio
#
# which results to the following prefix string when using the
# default template:
#
#  test-001_something_com.fabio
#
# The default is
#
# metrics.prefix = {{clean .Hostname}}.{{clean .Exec}}
```

### DIFF
--- a/fabio.properties
+++ b/fabio.properties
@@ -506,7 +506,8 @@
 #
 #    prefix.service.host.path.target-addr
 #
-# Available variables for prefix:
+# The value is expanded by the text/template package and provides
+# the following variables:
 #
 #  - Hostname:  the Hostname of the server
 #  - Exec:      the executable name of application
@@ -515,21 +516,22 @@
 #
 #  - clean:     lowercase value and replace '.' and ':' with '_'
 #
-# For example server hostname is:
+# Template may include regular string parts to customize final prefix
 #
-#  test-001.something.com
+# Example:
 #
-# When using the default template prefix will be as:
+#  Server hostname: test-001.something.com
+#  Binary executable name: fabio
+#
+#  The template variables are:
+#
+#  .Hostname =  test-001.something.com
+#  .Exec = fabio
+#
+# which results to the following prefix string when using the
+# default template:
 #
 #  test-001_something_com.fabio
-#
-#
-# At allow us add additional custom string Tags to any place of the prefix for customuzation:
-#
-# Template:   MyTag.{{clean .Hostname}}.{{clean .Exe}}
-#
-# Result:     MyTag.test-001_something_com.fabio
-#
 #
 # The default is
 #
--- a/fabio.properties
+++ b/fabio.properties
@@ -533,7 +533,7 @@
 #
 # The default is
 #
-# metrics.prefix = {{clean .Hostname}}.{{clean .Exe}}
+# metrics.prefix = {{clean .Hostname}}.{{clean .Exec}}
 
 
 # metrics.names configures the template for the route metric names.
--- a/fabio.properties
+++ b/fabio.properties
@@ -500,22 +500,40 @@
 # metrics.target =
 
 
-# metrics.prefix configures the prefix of all reported metrics.
+# metrics.prefix configures the template for the prefix of all reported metrics.
 #
 # Each metric has a unique name which is hard-coded to
 #
 #    prefix.service.host.path.target-addr
 #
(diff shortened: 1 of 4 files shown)


metrics/metrics.go
```diff
@@ -75,6 +78,33 @@ func NewRegistry(cfg config.Metrics) (r Registry, err error) {
...
+	if err != nil {
+		return "", err
+	}
+	host, err := hostname()
+	if err != nil {
+		return "", err
+	}
+	exe := filepath.Base(os.Args[0])
+
+	if err := t.Execute(&pfx, &prefixData{Hostname: host, Exec: exe}); err != nil {
```
magiconair: Pls make this an anonymous struct and drop the `prefixData` type , e.g.

```
b := new(bytes.Buffer)
var data struct { Hostname, Exec string}{ host, exe }
if err := t.Execute(b, &data); err != nil { ... }
return b.String(), nil
```

metrics/metrics.go
```diff
@@ -28,6 +28,9 @@ const DefaultNames = "{{clean .Service}}.{{clean .Host}}.{{clean .Path}}.{{clean
 // names stores the template for the route metric names.
 var names *template.Template
 
+// var pfx *template.Template
```
magiconair: pls add a proper comment

fabio.properties
```diff
@@ -500,17 +500,40 @@
...
 
-# metrics.prefix configures the prefix of all reported metrics.
+# metrics.prefix configures the template for the prefix of all reported metrics.
 #
 # Each metric has a unique name which is hard-coded to
 #
 #    prefix.service.host.path.target-addr
 #
-# When set to "default" the prefix is <hostname>.<executable>
+# Available variables for prefix:
```
magiconair: Can you make the comment similar in wording to the one for `metrics.names` ?

md2k:
i did required changes, hopefully it now much better.
and thanks for the tip about anonymous structures.

md2k:
and also i'm not sure if i should keep `metrics.prefix = default` processing for backward compatibility if someone who use Fabio and uncommented this option?
New functionality will broke his prefix immediately :)

magiconair:
@md2k I agree about the `default` behavior. Pls add this as an alias for the new template.

metrics/metrics.go
```diff
@@ -38,9 +41,9 @@ func init() {
 
 // NewRegistry creates a new metrics registry.
 func NewRegistry(cfg config.Metrics) (r Registry, err error) {
-	prefix := cfg.Prefix
-	if prefix == "default" {
-		prefix = defaultPrefix()
+
+	if prefix, err = parsePrefix(cfg.Prefix); err != nil {
```
magiconair: As discussed in the thread `default` should still work to avoid breaking existing behavior.

md2k:
Oki, i'll push changes tomorrow.

metrics/metrics.go
```diff
@@ -80,6 +83,10 @@ func NewRegistry(cfg config.Metrics) (r Registry, err error) {
 
 // parsePrefix parses the prefix metric template
 func parsePrefix(tmpl string) (string, error) {
+	// Backward compatibility condition for old metrics.prefix parameter 'default'
+	if tmpl == "default" {
```
magiconair: Would be nice to test this in `TestParsePrefix` to avoid regressions. Could you please add that and then I'll merge it.

md2k: ok, will do shortly

Issue #164, opened 2016-09-22
Routing by path


Salmondx:
Hi,
I have a question and didn't found a solution in docs.
Is it possible in fabio to pass to service only path variables after urlprefix?
E.g. I've got a service `test-service` with `urlpreifx-/test-service`.
And now i want to send http request like this:
`http localhost:9999/test-service/test/api`
I only want to pass `/test/api` part, not the whole path, and fabio does the second.

In nginx i can add `/` after a location path and all path variables will be propagated except one that used for routing.

magiconair:
Fabio does not support path stripping right now. There is a ticket open for this where you can read the reasoning. I'm probably going to add this in the future but I'm going to focus on other things first.

magiconair:
I'm closing this as a duplicate of #44. Change for path stripping is in progress. See branch.

PR #170, opened 2016-09-29
Load services by name and external nodes

Salmondx:
Hi,
For my purpose, i have to add some features to Fabio:
1. I've added a new boolean option called `registry.consul.register.byServiceName`. If it is set to `true`, then fabio will ignore `urlprefix-` tag existence and add service to routing table by its name. E.g. if service is called `test-service` then path to this service will looks like  `/test-service`. I've added this because we have lots of legacy services in consul and it's really difficult to change all deployment roles to add `urlprefix-` in consul. Such option provide us a possibility to migrate to fabio without pain.
2. Also i've added a flag option called `registry.consul.service.externalNodes`.  This option accepts a list of external nodes in consul (https://www.consul.io/docs/guides/external.html) separated by `,`. If this option exists, then fabio will register all services from this nodes to it's routing table. Purpose of this option is to have external services in our routing table and standard fabio way is not acceptable cause external services doesn't have healthchecks. Comma-separated flag is also a work-around cause there is no possibility in Consul (0.7.0) API to tag node as external.
3. And another option, called `proxy.stripPath`. There were a lot of issues about this problem (e.g. my last one #164). If this option is enabled, then path to service will be stripped (E.g. `GET /test-service/api` → `GET /api`


magiconair:
@Salmondx Thanks for the contribution. A couple of comments:

1. That sounds like an interesting feature but I would prefer a different implementation. Right now the code that gets the services from consul and the code that builds the routing table from this service list is one. What would make sense is to separate fetching services and generating a routing table from a list of services into separate pieces to make both of them configurable independently. So in essence build an internal plugin like the `Picker` or `Matcher` that builds a routing table from a list of services based on some config. Then your approach for your legacy services would just be another filter. Or we could have multiple as long as they are exclusive.

2. I think you can solve the external nodes problem by adding static routes to consul. Just add the routes you need. This is kind of a limitation of supporting only one service registry at a time. If you could have both consul and file then you can just copy a file somewhere. 

3. I'm already working on the path stripping. 

Taking these things into account I won't merge the change as is. Also, because it combines several features into one larger change. If the patch is working for you right now feel free to use it and I can refactor the code as described so that you can build a filter. Or you can provide some code if you want. Separate PR please.

Salmondx:
@magiconair Thanks for review!
1. Ok, i'll fix this and make a separate PR.
2. Hm, static route is not a solution for me. For example, in my company we register legacy soap services as services on external node. So, node ```test``` has lots of services and its not possible to describe them in external config file, especially when fabio doesn't support dynamic config reload.
May be there are another solution for my problem?
3. Ok, it'll be really cool feature!

magiconair:
As for 2. why don't you register them in consul on deployment or write a small tool which manages the registration? That seems like the Right Thing (TM) to do instead of patching the load balancer. This would also solve your first problem.

magiconair:
So please wait with another PR until we've settled on an approach.

Salmondx:
@magiconair they are registered in consul on external node.
E.g. ```soap-services``` node has several registered services. But there is no possibility for us to register healthcheck cause they doesn't provide such endpoint.

magiconair:
Can't you just register a TCP check or a simple script that performs a function that concludes that the service is up? You don't need to have a specialized `/health` endpoint to perform the health check.

magiconair:
Sorry, missed the fact that there are several services on a single node. So the underlying problem is that fabio is looking for `passing` services in consul but your services do not have a health check. fabio has an option to specify the consul service status but that may be too broad. `registry.consul.service.status`. You could set this to `passing,unknown`

Salmondx:
I cant, cause consul api doesn't return services without registered health checks through ```/v1/health/state``` endpoint that used in fabio.

magiconair:
How do you know your service is up then?

Salmondx:
It's a third party services and we have no control over them and no fallback is possible. Like in consul doc's, external node points on google service :)

magiconair:
Hmm, this registers an external service on my local consul including tags and fabio integration. What am I missing?

```
# register the service
$ curl -i -XPOST -H 'Content-Type: application/json' -d '{"ID": "marktplaats", "Name": "marktplaats", "Tags": ["urlprefix-www.marktplaats.nl/"], "Address": "91.195.49.34", "Port": 80, "EnableTagOverride": false, "Check": {"HTTP": "http://91.195.49.34/", "Interval": "15s"} }' 'http://localhost:8500/v1/agent/service/register'
HTTP/1.1 200 OK

# check that health check is passing
$ curl -i 'http://localhost:8500/v1/health/state/passing?pretty'
HTTP/1.1 200 OK
Content-Type: application/json
X-Consul-Index: 63
X-Consul-Knownleader: true
X-Consul-Lastcontact: 0
Date: Fri, 30 Sep 2016 11:26:26 GMT
Transfer-Encoding: chunked

[
    {
        "Node": "LM-AMS-00963965",
        "CheckID": "serfHealth",
        "Name": "Serf Health Status",
        "Status": "passing",
        "Notes": "",
        "Output": "Agent alive and reachable",
        "ServiceID": "",
        "ServiceName": "",
        "CreateIndex": 4,
        "ModifyIndex": 4
    },
    {
        "Node": "LM-AMS-00963965",
        "CheckID": "service:marktplaats",
        "Name": "Service 'marktplaats' check",
        "Status": "passing",
        "Notes": "",
        "Output": "HTTP GET http://91.195.49.34/: 200 OK Output:   ..."
    }
]

# check that fabio picks it up
$ fabio

...
2016/09/30 13:27:31 [INFO] Updated config to
route add marktplaats www.marktplaats.nl/ http://91.195.49.34:80/ tags "urlprefix-www.marktplaats.nl/"
...

# test that it works (separate terminal)
curl -i -H 'Host: www.marktplaats.nl' 'localhost:9999/'

# deregister service
curl -i -XPOST http://localhost:8500/v1/agent/service/deregister/marktplaats
```

Salmondx:
With service you also register a healthcheck, and in my case its not an option:
```
"Check": {"HTTP": "http://91.195.49.34/", "Interval": "15s"} 
```

magiconair:
If Fabio is supposed to route HTTP traffic to this service how can this not be an option? You need to have some monitoring around this which verifies that your external service is up. If you don't do this you can register a health check that never fails like shell with /bin/true. 

I'm not saying you don't have a case but right now I fail to see it. Please keep explaining what I'm missing.

Salmondx:
Because a fake healthcheck is a work around, and external services shouldn't be monitored cause you can't change anything if they fail. For example, if i have a google as external service, i should also use an http health check and add unnecessary load on my consul cluster? May be i'm wrong.

magiconair:
The fake health check would indeed be just a workaround. 

The hole in your argument is that you claim that you can reach the service via HTTP since fabio should forward traffic to it but you cannot use exactly the same mechanism to establish the consul health check. That doesn't make sense to me.

magiconair:
Hmm, I should have read your full response instead of just the first sentence. :/

OK, so in essence you could monitor the external service but you don't want to since you don't think this is the right thing to do. You also do not want to provide a always-true health check to ensure that fabio works in your environment since you consider that a kludge. You would rather change the load balancer instead? Is this your argument?

I get your point but I'm not sure I agree with it. You are basically saying that health checks should be actionable which is fine. But it falls short if your service depends on the availability of the external service. Then you **want** to know that your service is either failing or degraded because of an external factor even if you cannot change it. At least you can stop digging and maybe escalate the problem. This then is the action of your failing health check.

From that perspective I'd absolutely would check external services if my own service depends on them.

And by doing that it solves your first two problems at once without any code change which has to be maintained. Only the path stripping is missing.

magiconair:
@Salmondx Are you still using this patch? Path stripping has been added some time ago.

magiconair:
Closing because of age. Pls comment to re-open.

PR #180, opened 2016-10-17
added glob host matching - fixes #163

### LINKED ISSUE #163: Support glob host matching
Apologies if there is a feature for this already, I wasn't able to find it.

I was curious if consul can support wildcard domains in the urlprefix.

```
E.g. *.mydomain.com/route
Or .mydomain.com/route
Or .*\.mydomain.com\/route
```

Or something similar to that effect?

discobean:
Hi, I've added some code to support a fix for #163. - Pre warning this is my first bit of golang.

I updated the lookup() method in the Table class, that:
1. If no host defined, searches just paths (existing functionality)
2. If a host is defined then does a direct match (existing functionality)
3. If neither, then does a glob search against hosts, then does path matching for each host

I wasn't sure the best way to implement it, but this seems to work

Considering it won't affect existing functionality, I didn't add any additional configuration options, and 'go test' passed.

Also since it's my first attempt at go, I'm not sure whether it is the 'best (go) way of doing' things, or that my function and variable names are right.

### DIFF
--- a/route/table.go
+++ b/route/table.go
@@ -8,7 +8,7 @@ import (
 	"net/url"
 	"sort"
 	"strings"
-	"regexp"
+	"path"
 	"sync"
 	"sync/atomic"
 
@@ -320,15 +320,17 @@ func (t Table) lookup(host, lookupPath, trace string) *Target {
 		return t.matchHostPath(host, lookupPath, trace)
 	}
 
-	/* // TODO if no regex configuration, just return this
+/*
+	// TODO if no regex configuration, just return this
 	// If the host is defined, run a search against that exact host name
 	var foundTarget = t.matchHostPath(host, lookupPath, trace)
 	if foundTarget != nil {
 		return foundTarget
 	}
-	*/
+*/
 
 	// TODO give an option for regex or glob host configuration
+	// think about how to do regex whether to inject ^$ etc.
 
 	// If the user wants to run regex search (TODO, do regex and make configuration)
 	// Create a new array of targets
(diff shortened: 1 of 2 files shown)


magiconair:
Cool. I'll have a look later

discobean:
Thanks, I also think the iterating in the glob matching would slow it down with many hosts and some sort of index cache would have been good, but just didn't have enough time on this first iteration.

discobean:
I can foresee some problems when matching routes based on order - because I think in the code the ordering is based on the path, not the domain.

So ideally the code should find all the matching hosts and paths, and return the most restrictive result.

discobean:
I think there is more to it also, say you have 2 routes:

1. urlprefix-*.abc.com/myurl
2. urlprefix-foo.abc.com/myurl

A request to foo.abc.com/myurl would need to match against route 2, my hacky code doesn't take this into account.

discobean:
Yep that was the original problem I had, totally agree it can be done better, but I didn't want to dig too much into it without more experience with go.  I'll look into the code a bit more tomorrow.

osigida:
hi all,

the feature looks interesting, any progress on it?

discobean:
@osigida, unfortunately no, my fork is actually working but it is not the best way of doing it, so it needs a rewrite

magiconair:
Let me have another look

Frank Schröder

magiconair:
I took a stab at this in #163 Please have a look.

magiconair:
I'm going to close this PR since a change for #163 has been merged to master. Please open new tickets with a reference to #163 if there are issues with the provided implementation.

discobean:
Thanks @magiconair

PR #190, opened 2016-11-21
allowing route to match host running at any port aimed url

abhishekkr:
Signed-off-by: AbhishekKr <abhikumar163@gmail.com>

Hey, as per our conversation today
https://twitter.com/magiconair/status/800626305490259968

I've made few changes to Fabio code to support request hostname provided with any port, not just `80|443`.
This works for my use-case without me to provide `Host:` header in `urlprefix-host/` scenario.

### DIFF
--- a/route/table.go
+++ b/route/table.go
@@ -229,13 +229,11 @@ func (t Table) route(host, path string) *Route {
 // and removes the default port if present.
 func normalizeHost(req *http.Request) string {
 	host := strings.ToLower(req.Host)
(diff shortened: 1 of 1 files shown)


magiconair:
That code is there for a different reason. Could you please explain the problem you are trying to solve? This isn't clear from the Twitter communication. 

My guess is that you just need to add the port to the host in the urlprefix tag but I'd need to see the details. 

Thx
Frank Schröder

abhishekkr:
@magiconair ah, now I get it... wrong PR, closing it

Issue #227, opened 2017-02-03
Q: Why does Fabio log the entire route table on every change?


stevehorsfield:
### Context

Based on: https://github.com/eBay/fabio/blob/128b59ccb543832312ea2050b8154b4fd0ee93b8/route/table.go#L45-L54

We see thousands of messages like:

```
2017/02/03 14:18:19 [INFO] Updated config to
```

and

```
route add my-service my-service/ http://a.b.c.d:port/ tags "private,private:my-service.service.consul/,private:my-service/"
```

### Expected behaviour

We expect to only see logs of changes (the delta), not the entire route table.

### Perceived behaviour

I think that Fabio is logging the entire route table every time a change occurs.

### Comment

Could you please confirm if this is expected behaviour? If so, should there be an option to disable the output of these messages? If not, can we get it fixed? We're happy to put a PR together.

A delta would certainly be a desirable output.

magiconair:
At this point this is the expected behavior. I agree that only (or additionally) logging the delta may be beneficial for debugging purposes and I may do that in the future. As stated in #226 the reason you are seeing thousands of messages of the routing table is because something in your environment is not stable. That is the only information you can derive from that. Once you find and fix that problem then you will only see the updated routing table. I'll add the delta thing to the list of things to add. This would indeed be useful.

stevehorsfield:
We have a lot of services, some of which do flap (particularly in pre-prod). However, I'm pretty sure our logging isn't due to flapping per se (it may be due to the way Consul is handling health check changes). In any case, if we can't get this output reduced through some mechanism we'll have to fork. Another option is to control the logging detail through feature flags. Is that something you are open to?

magiconair:
If this is a production issue which you can't fix easily them I'm open to bump the priority on this. Usually, I push back on adding features which mask a different problem. For now just comment out the log lines and build it yourself. I'll see what I can do.

magiconair:
The other thing we're doing in QA environments is to run fabio via upstart and rotate the log files once they've reached a certain size.

stevehorsfield:
We're shipping our logs directly to a SaaS log vendor, so rotation isn't needed for us. I'm looking at Consul to see if I can find the logic that might link to this behaviour. Appreciate your feedback.

magiconair:
I think we can close this one. Please comment if there is something to add and I'll re-open if necessary.

Issue #226, opened 2017-02-03
Q: Does Fabio handle Consul service health changes well?


stevehorsfield:
### Context

Based on https://github.com/eBay/fabio/blob/128b59ccb543832312ea2050b8154b4fd0ee93b8/registry/consul/service.go#L20-L30.

We see a huge number of log messages such as:
```
2017/02/03 12:13:53 [INFO] consul: Health changed to #41083871
```

### Expected behaviour:

Log messages only when health state of a service is changed.

### Perceived behaviour

Log messages any time the health data is written.

My gut feeling is that Fabio is acting on any change to the Consul health value, not only on a change in state (i.e. critical/warning/healthy). Could you please confirm.

### Comment

Could you please confirm if this is expected behaviour? If so, should there be an option to disable the output of these messages? If not, can we get it fixed? We're happy to put a PR together.

magiconair:
If you see a lot of these messages then you have a flapping service in your setup. This message is printed when the internal state of the consul cluster changes which happens on registration or deregistration of a service or when the health endpoint of a service is flapping. My suggestion is to run `consul monitor --log-level=debug` and see what is going on.

magiconair:
Oh, and to answer your question: Yes, fabio acts on *any* change in the consul state, pulls the full state to build the routing table. The assumption is that your setup is stable once it has converged and that there isn't a lot of change. Hence, if you see a lot of change your setup isn't stable and hasn't converged.

stevehorsfield:
This sounds wrong to me, but perhaps I am missing something. Why would you consider an index change to be equivalent to a state change?

magiconair:
Since this is the way the consul and it's blocking API works. consul maintains a replicated state among its master nodes. This state is indexed with a number - the version. Every time I make a query I get the version that result is based on as meta data in the response so that I can use that for subsequent queries, e.g. please notify me if your state changes beyond version X. As long as the state does not change (no service deployment, health change) fabio just performs a blocking wait (long poll) on the consul API. Once the state changes fabio pulls the *full* state, e.g. all services, all health checks to determine the new routing table. If the routing table did not change then fabio won't do anything to the main proxy thread but it will report that the consul state has changed nevertheless. This is an indication that something is constantly changing in the consul state, e.g. a service which doesn't have a `urlprefix-` tag.

I may not need to report a change every time I get notified of one but it also provides an indication that there is something else going on in your system. fabio could also identify the delta and log it.

stevehorsfield:
[...]

PR #236, opened 2017-02-07
Adding mdllog log package and also changing fabio code

avichalbadaya:
Includes a new package mdllog for logging which is a
wrapper on top of  go log package. Also made changes
in fabio code to include new package and enable log
level setting using config argument.

### DIFF
--- a/admin/api/api.go
+++ b/admin/api/api.go
@@ -3,8 +3,8 @@ package api
 
 import (
 	"encoding/json"
-	"log"
 	"net/http"
+	"github.com/eBay/fabio/mdllog"
 )
 
 func writeJSON(w http.ResponseWriter, r *http.Request, v interface{}) {
@@ -19,7 +19,7 @@ func writeJSON(w http.ResponseWriter, r *http.Request, v interface{}) {
 	}
 
 	if err != nil {
-		log.Print("[ERROR] ", err)
+		mdllog.Error.Print("[ERROR] ", err)
 		http.Error(w, "internal error", 500)
 		return
 	}
--- a/admin/api/manual.go
+++ b/admin/api/manual.go
@@ -2,10 +2,10 @@ package api
 
 import (
 	"encoding/json"
-	"log"
 	"net/http"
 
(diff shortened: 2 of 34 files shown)


magiconair:
First, let me thank you for the PR even though you've closed it. Generally, I prefer to start larger changes with a discussion as outlined in https://github.com/eBay/fabio/wiki/Contributing but I should make this more visible in the README, I guess. It is somewhat hidden at the bottom of the Wiki.

As for the change itself, the reason I didn't choose to use a leveled log package like glog or logrus is simplicity. The default log package and the level prefix are good enough for what fabio needs to do. Therefore, this is one less dependency to worry about. Also, log level filtering has the potential that people want to tune out excessive logging instead of finding the root cause. Issue #226 is such an example where investigating the excessive logging lead to the discovery of volatile health checks. 

Our production service which handles about 5.000 req/sec generates 15.000 log lines or 1 MB per day with most of the log messages about routes that don't exist. Once we address that, we should have about 100KB of logs per day. Therefore, IMHO fabio does not log much and if it does you should find out why and not suppress it. 

Frank

magiconair:
Since this came up in #227 as well I think it could make sense to just log the delta instead of the full route table. I'm not a fan of log filtering for the reasons I've outlined. If the logging of the full routing table is your main concern then lets work on that. Also, if you want to do log filtering then there are easier ways to do this without the `mdllog` package which doesn't change code all over the place.

avichalbadaya:
@magiconair : Yes, I am glad to hear that. If we can log just delta instead of full route table, it would certainly solve our issue with log volume. We wouldn't need log filtering in that case.

killcity:
+1

avichalbadaya:
Hi Frank,

Thanks much creating #238 . I also wanted to ask on point you mentioned about easier ways of log filtering. Can you share some details on what ideas you have to achieve it without much change in code ?

magiconair:
I need to dig this up but I think the general approach is to put a filter writer in the log output stream which allows only strings with a certain prefix. Lets first see whether #238 fixes your problem.

PR #246, opened 2017-02-26
added log export to graylog with gelf writer

raben2:
Hi Frank, 
i wrote a log exporter for Graylog2 in Gelf format. 

You can test it using this graylog docker-compose https://github.com/raben2/graylog-docker-cluster

Best Regards Georg

### DIFF
--- a/config/default.go
+++ b/config/default.go
@@ -60,4 +60,11 @@ var defaultConfig = &Config{
 			APIApp: "fabio",
 		},
 	},
+	LogServer: LogServer{
+		Enabled: false,
+		Protocol: "gelf",
+		Transport: "udp",
+		Port:      "12201",
+		Address:   "127.0.0.1",
+	},
 }
--- a/config/load.go
+++ b/config/load.go
@@ -164,6 +164,11 @@ func load(cmdline, environ, envprefix []string, props *properties.Properties) (c
 	f.StringVar(&cfg.UI.Addr, "ui.addr", defaultConfig.UI.Addr, "address the UI/API is listening on")
 	f.StringVar(&cfg.UI.Color, "ui.color", defaultConfig.UI.Color, "background color of the UI")
 	f.StringVar(&cfg.UI.Title, "ui.title", defaultConfig.UI.Title, "optional title for the UI")
+	f.BoolVar(&cfg.LogServer.Enabled, "logserver.enabled", defaultConfig.LogServer.Enabled, "enable external logging")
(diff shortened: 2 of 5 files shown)


magiconair:
How about a generic `log.target` parameter that can take any number of log targets? 

`log.target = stderr,gelf[;addr=host:port][;proto=udp]`

To implement this you could use a similar approach as for the cert source and use `f.KVSliceVar()` for parsing. However, the current `KVSliceVar()` code requires all values to have a name. You would then have to format the values as

`log.target = target=stderr,target=gelf, ...`

This feels like stuttering and it would make sense to extend the `KVSliceVar()` method with a prefix which is automatically added to all values before parsing. So you would call `KVSliceVar("target=", ...`.

This would also make sense for the `metrics.target` parameter.

Also, you need to have tests for all new command line parameters.

raben2:
Sure great idea, i will look into that. 
 
Am i wrong or is the travis build failing because you have to reconfigure the dependencies?

magiconair:
Since a number of people are showing interest in #80 we should try to come up with a unified approach to cover both. @raben2 are you still working on this?

raben2:
Yeah sure, i will try to look at it asap. Right now i am doing a very important vacation ;)

magiconair:
@raben2 Then this can wait. Vacation is more important. Ping me when you're back.

magiconair:
@raben2 if your patch includes the changes from master then the rebasing didn't work properly. Could you try again, please? If your change is small enough then just squash it and force push. I usually don't care how one got there and look only at the end result. 

Make sure the vendored lib is in a separate commit. Sometimes I put the refactoring that is necessary to implement a feature into a separate commit. (e.g. replacing all `log` calls with some custom interface) to separate the refactoring from the actual feature.

magiconair:
@raben2 I've just pushed a change to the #80 branch to rename the logging parameters. I've moved them to the global level with `log.*` My suggestion for this PR is this:

```
# these should be the default values
log.graylog.protocol = gelf
log.graylog.transport = udp
log.graylog.addr = 127.0.0.1:12201

# these should be optional
log.error.target = graylog
log.access.target = graylog
```

raben2:
Oh great thanks for adding this. I try to fix up the pr ASAP

magiconair:
This is on master now

magiconair:
@raben2 Are you still working on this?

nanoz:
This looks awesome, can't wait for this feature to hit a release !

magiconair:
@raben2 ping

nanoz:
Nooooooo !!! :)

Issue #181, opened 2016-10-28
Add HTTPS upstream support


ak66982:
Hi,

Is there any way to register https:// backends without using tcp+sni, with fabio performing ssl termination on the client side, and talking https to the backends (re-encoding traffic) ?  The reason for this is that frontend certificate is not the same as the ones that are used internally, so plain tcp+sni will not work.

Regards,
Alex

magiconair:
fabio does not support that yet. I'll probably add that at a later stage when I'm addressing #111

Narrowbeam:
Would be really good to see this - its the only feature stopping our adoption of this project.  Fingers crossed you can add it fairly soon.

magiconair:
@Narrowbeam It is on the list. I've spent some time refactoring the listener and proxy code in the `release-branch-1.4` to make this simpler. The addition of the TCP proxy was already a step in that direction. So stay tuned. :)

shadowfax-chc:
@magiconair is this as simple as adding a new conditional in [registery/consul/service.go](https://github.com/eBay/fabio/blob/master/registry/consul/service.go#L116-L119)?

Not sure what kind of configuration you might want around it.

magiconair:
It could be. However, I think that you need to have some cert management options as well since upstream certs are probably self-signed. There may be more but I think that would be a good start. 

Frank Schröder

magiconair:
Thanks @shadowfax-chc for implementing this. This is great progress. I'll release 1.4.2 soon with this.

PR #257, opened 2017-04-04
Add https upstream support

shadowfax-chc:
This seems to work for me. I generally have a internal root CA that is trusted via standard system certs. The `skipverify` option could be used to deal with self-signed certs, but would obviously skip verification.

I was also thinking of maybe having another certificate store for CAs that could then be selected via a tag option. If that seems like a good idea I can make another PR or amend this one.

Ref: #181

### DIFF
--- a/main.go
+++ b/main.go
@@ -117,16 +117,22 @@ func newHTTPProxy(cfg *config.Config) http.Handler {
 	log.Printf("[INFO] Using routing strategy %q", cfg.Proxy.Strategy)
 	log.Printf("[INFO] Using route matching %q", cfg.Proxy.Matcher)
 
-	return &proxy.HTTPProxy{
-		Config: cfg.Proxy,
-		Transport: &http.Transport{
+	newTransport := func(tlscfg *tls.Config) *http.Transport {
+		return &http.Transport{
 			ResponseHeaderTimeout: cfg.Proxy.ResponseHeaderTimeout,
 			MaxIdleConnsPerHost:   cfg.Proxy.MaxConn,
 			Dial: (&net.Dialer{
 				Timeout:   cfg.Proxy.DialTimeout,
 				KeepAlive: cfg.Proxy.KeepAliveTimeout,
 			}).Dial,
-		},
+			TLSClientConfig: tlscfg,
+		}
+	}
+
+	return &proxy.HTTPProxy{
+		Config:            cfg.Proxy,
+		Transport:         newTransport(nil),
+		InsecureTransport: newTransport(&tls.Config{InsecureSkipVerify: true}),
 		Lookup: func(r *http.Request) *route.Target {
 			t := route.GetTable().Lookup(r, r.Header.Get("trace"), pick, match)
 			if t == nil {
--- a/proxy/http_proxy.go
+++ b/proxy/http_proxy.go
@@ -27,6 +27,11 @@ type HTTPProxy struct {
 	// The proxy will panic if this value is nil.
 	Transport http.RoundTripper
 
+	// InsecureTransport is the http connection pool configured with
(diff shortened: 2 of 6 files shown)


magiconair:
@shadowfax-chc Thanks for working on this but this isn't correct.

There are two TCP connections in play here: one inbound and one outbound/upstream. The listener is the inbound TCP connection and the proxy is the outbound TCP connection. To allow outbound `https` connections you only need to set the protocol in the target URL which your first patch accomplishes. The Go std library should do the rest. 

This requires however that the certificates can be validated through the default root CAs as configured on the system. If you want to allow certificate verification with a custom certificate source then this must be configured in the `route` command/`urlprefix-` tag.

Lets start with supporting `skipVerify` on the outbound connection since this is rather simple and should solve most of the problems. Supporting custom root CAs would require to either override the `http.Client` that is used by the reverse proxy (which isn't possible ATM) or use a custom transport per request (which defeats connection caching). This most likely requires a somewhat larger refactor. 

To support `skipVerify` you need to do several things:

1. Add a `TLSSkipVerify` bool flag to the `route.Target` and set it in the `route.addTarget` method in `route/route.go`
1. Create another `http.Transport` which has the `InsecureSkipVerify` flag set to `true` and store it in the `HTTPProxy`.
1. Pick which transport to use in the `HTTPProxy.ServeHTTP` method based on the flag in the selected target.

Once we find a way to have a custom client/transport per request this solution can easily be refactored.

You don't have to (and shouldn't) configure the `Listener` since this is for the inbound connection. `skipVerify` may be useful here as well but that's a different thing. 

Also, please provide an integration test for both cases in `proxy/http_integration_test.go`.

shadowfax-chc:
Ah, that makes sense. Let me make sure I am setting this option in the correct place now, and I will get some tests.

route/route.go
```diff
@@ -70,6 +71,13 @@ func (r *Route) addTarget(service string, targetURL *url.URL, fixedWeight float6
 	}
 	if r.Opts != nil {
 		t.StripPath = r.Opts["strip"]
+
+		tlsskipverify := false
```
magiconair: This is enough since this is also in line with the option parser in `config`

```go
t.TLSSkipVerify := r.Opts["skipverify"] == "true"
```

proxy/http_proxy.go
```diff
@@ -20,6 +20,8 @@ type HTTPProxy struct {
 	// The proxy will panic if this value is nil.
 	Transport http.RoundTripper
 
+	InsecureTransport http.RoundTripper
```
magiconair: Pls add a comment.

route/target.go
```diff
@@ -32,4 +32,7 @@ type Target struct {
 
 	// timerName is the name of the timer in the metrics registry
 	timerName string
+
```
magiconair: "TLSSkipVerify disables certificate validation for upstream TLS connections."

Could you also move the variable below the `StripPath` variable, please?

main.go
```diff
@@ -101,6 +101,17 @@ func newHTTPProxy(cfg *config.Config) http.Handler {
 				KeepAlive: cfg.Proxy.KeepAliveTimeout,
 			}).Dial,
 		},
+		InsecureTransport: &http.Transport{
```
magiconair: Maybe move this into a local `newTransport := func(tlscfg *tls.Config) http.Transport` func to avoid the duplication.

magiconair:
Oh, I can actually change the files myself now. Interesting. Didn't know that.

magiconair:
I'll try that on the next PR, though.

shadowfax-chc:
@magiconair made the requested changes and added tests

magiconair:
LGTM. Merged it. Thank you!

PR #268, opened 2017-04-21
Add support for TLSSkipVerify for https consul fabio check

Ginja:
When serving the UI/API over HTTPS, the fabio consul health check will most likely fail as it's trying to connect to the page over its IP. Most certificates don't include IP SANs. This flag allows users to toggle whether or not to skip TLS verification for this particular check.

Edit: This will require [Consul >= 0.7.2](https://github.com/hashicorp/consul/blob/master/CHANGELOG.md#072-december-19-2016)
hashicorp/consul#1984
hashicorp/consul#2530

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -106,17 +106,18 @@ type File struct {
 }
 
 type Consul struct {
-	Addr          string
-	Scheme        string
-	Token         string
-	KVPath        string
-	TagPrefix     string
-	Register      bool
-	ServiceAddr   string
-	ServiceName   string
-	ServiceTags   []string
-	ServiceStatus []string
-	CheckInterval time.Duration
-	CheckTimeout  time.Duration
-	CheckScheme   string
+	Addr               string
+	Scheme             string
+	Token              string
+	KVPath             string
+	TagPrefix          string
+	Register           bool
+	ServiceAddr        string
+	ServiceName        string
+	ServiceTags        []string
+	ServiceStatus      []string
+	CheckInterval      time.Duration
+	CheckTimeout       time.Duration
+	CheckScheme        string
+	CheckTLSSkipVerify bool
 }
--- a/config/load.go
+++ b/config/load.go
@@ -162,6 +162,7 @@ func load(cmdline, environ, envprefix []string, props *properties.Properties) (c
 	f.StringSliceVar(&cfg.Registry.Consul.ServiceStatus, "registry.consul.service.status", defaultConfig.Registry.Consul.ServiceStatus, "valid service status values")
 	f.DurationVar(&cfg.Registry.Consul.CheckInterval, "registry.consul.register.checkInterval", defaultConfig.Registry.Consul.CheckInterval, "service check interval")
 	f.DurationVar(&cfg.Registry.Consul.CheckTimeout, "registry.consul.register.checkTimeout", defaultConfig.Registry.Consul.CheckTimeout, "service check timeout")
+	f.BoolVar(&cfg.Registry.Consul.CheckTLSSkipVerify, "registry.consul.register.checkTLSSkipVerify", defaultConfig.Registry.Consul.CheckTLSSkipVerify, "service check TLS verifcation")
 	f.IntVar(&cfg.Runtime.GOGC, "runtime.gogc", defaultConfig.Runtime.GOGC, "sets runtime.GOGC")
 	f.IntVar(&cfg.Runtime.GOMAXPROCS, "runtime.gomaxprocs", defaultConfig.Runtime.GOMAXPROCS, "sets runtime.GOMAXPROCS")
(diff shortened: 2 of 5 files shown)


config/default.go
```diff
@@ -43,17 +43,18 @@ var defaultConfig = &Config{
...
-			CheckTimeout:  3 * time.Second,
-			CheckScheme:   "http",
+			Addr:                 "localhost:8500",
+			Scheme:               "http",
+			KVPath:               "/fabio/config",
+			TagPrefix:            "urlprefix-",
+			Register:             true,
+			ServiceAddr:          ":9998",
+			ServiceName:          "fabio",
+			ServiceTLSSkipVerify: false,
```
magiconair: You can omit that since this is the default

config/load_test.go
```diff
@@ -426,6 +426,13 @@ func TestLoad(t *testing.T) {
 			},
 		},
 		{
+			args: []string{"-registry.consul.register.tlsskipverify=false"},
```
magiconair: pls test with `true` since `false` is the default

config/config.go
```diff
@@ -106,17 +106,18 @@ type File struct {
...
-	CheckScheme   string
+	Addr                 string
+	Scheme               string
+	Token                string
+	KVPath               string
+	TagPrefix            string
+	Register             bool
+	ServiceAddr          string
+	ServiceName          string
+	ServiceTLSSkipVerify bool
```
magiconair: This should be `CheckTLSSkipVerify` and pls move it to the bottom below `CheckScheme`

config/load.go
```diff
@@ -158,6 +158,7 @@ func load(cmdline, environ, envprefix []string, props *properties.Properties) (c
 	f.BoolVar(&cfg.Registry.Consul.Register, "registry.consul.register.enabled", defaultConfig.Registry.Consul.Register, "register fabio in consul")
 	f.StringVar(&cfg.Registry.Consul.ServiceAddr, "registry.consul.register.addr", defaultConfig.Registry.Consul.ServiceAddr, "service registration address")
 	f.StringVar(&cfg.Registry.Consul.ServiceName, "registry.consul.register.name", defaultConfig.Registry.Consul.ServiceName, "service registration name")
+	f.BoolVar(&cfg.Registry.Consul.ServiceTLSSkipVerify, "registry.consul.register.tlsskipverify", defaultConfig.Registry.Consul.ServiceTLSSkipVerify, "service tls verification")
```
magiconair: pls change to `registry.consul.register.checkTLSSkipVerify` since I've - unfortunately - used camelCase with the other `check` parameters as well. The comment should be `service check TLS verifcation`

Ginja:
@magiconair Thanks for the review! Made the changes, and squashed everything into one commit.

config/default.go
```diff
@@ -43,17 +43,18 @@ var defaultConfig = &Config{
...
+			KVPath:             "/fabio/config",
+			TagPrefix:          "urlprefix-",
+			Register:           true,
+			ServiceAddr:        ":9998",
+			ServiceName:        "fabio",
+			ServiceStatus:      []string{"passing"},
+			CheckInterval:      time.Second,
+			CheckTimeout:       3 * time.Second,
+			CheckScheme:        "http",
+			CheckTLSSkipVerify: false,
```
magiconair: pls remove this since `false` is the default

Ginja: It is, but then how do I expose the parameter? 

https://github.com/fabiolb/fabio/pull/268/files#diff-f1ea9ed43b0b0ed74a6cf0f71dad3196R165

Do you want `defaultConfig.Registry.Consul.CheckTLSSkipVerify` just to be `false` in the link above?

magiconair: The field itself is defined in `config.Config` and not here. The value that you are setting in `defaultConfig` is automatically set by the compiler and therefore redundant. `false` is the default value for any `bool` variable.

magiconair:
`fabio.properties` also needs to be updated. Could you please add that?

config/default.go
Ginja: Oh okay, thanks for the info! So just to clarify, should I remove this line entirely:

`f.BoolVar(&cfg.Registry.Consul.CheckTLSSkipVerify, "registry.consul.register.checkTLSSkipVerify", defaultConfig.Registry.Consul.CheckTLSSkipVerify, "service check TLS verifcation")`

or set it to:

`f.BoolVar(&cfg.Registry.Consul.CheckTLSSkipVerify, "registry.consul.register.checkTLSSkipVerify", false, "service check TLS verifcation")`

or remove it completely?

And I'll update fabio.properties in my next push. Sorry about that.

magiconair:
You should remove the line from the defaultConfig only. Leave the rest. 

Frank Schröder

Ginja:
Changes made.

magiconair:
Made some minor text changes in the `fabio.properties` and then merged it. Thanks for providing this patch! :)

PR #273, opened 2017-04-26
Use scheme() throughout addHeaders

hkolk:
Noticed while adding some new headers that there was a bit of duplicated code. To avoid mangling everything into one pull request, I made a separate one.

Tests were running ok on this change:
```
cloud@fabio:~/go/src/github.com/fabiolb/fabio/proxy$ go test
2017/04/26 13:18:08 [INFO] cert: Store has certificates for ["example.com"]
PASS
ok  	github.com/fabiolb/fabio/proxy	0.576s
```


magiconair:
Good catch.

magiconair:
Merged

PR #292, opened 2017-05-12
Add unique request id

bkmit:
This patch add proxy.header.uniquerequestid configuration option for
specifying header name for adding unique ID to each http request proxied by
fabio.
  This header value can be logged in fabio access log and used for request
processing tracking across different applications.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -57,6 +57,7 @@ type Proxy struct {
 	TLSHeader             string
 	TLSHeaderValue        string
 	GZIPContentTypes      *regexp.Regexp
+	RequestID             string
 }
 
 type Runtime struct {
--- a/fabio.properties
+++ b/fabio.properties
@@ -353,6 +353,15 @@
 # proxy.header.tls.value =
 
 
+# proxy.header.requestid configures the header for the adding a unique request id.
+# When set non-empty value the proxy will set this header on every request to the
+# unique UUID value.
+#
+# The default is
+#
+# proxy.header.requestid =
+
+
 # proxy.gzip.contenttype configures which responses should be compressed.
 #
 # By default, responses sent to the client are not compressed even if the
--- a/NOTICES.txt
+++ b/NOTICES.txt
@@ -146,3 +146,9 @@ golang.org/go
 https://github.com/golang/go
 License: BSD 3-clause (https://github.com/golang/go/LICENSE)
 Copyright (c) 2009 The Go Authors. All rights reserved.
+
+
+github.com/rogpeppe/fastuuid
+https://github.com/rogpeppe/fastuuid.git
+License: BSD 3-clause (https://github.com/google/uuid/LICENSE)
+Copyright © 2014, Roger Peppe All rights reserved.
--- a/config/load.go
+++ b/config/load.go
@@ -126,6 +126,7 @@ func load(cmdline, environ, envprefix []string, props *properties.Properties) (c
 	f.StringVar(&cfg.Proxy.ClientIPHeader, "proxy.header.clientip", defaultConfig.Proxy.ClientIPHeader, "header for the request ip")
 	f.StringVar(&cfg.Proxy.TLSHeader, "proxy.header.tls", defaultConfig.Proxy.TLSHeader, "header for TLS connections")
 	f.StringVar(&cfg.Proxy.TLSHeaderValue, "proxy.header.tls.value", defaultConfig.Proxy.TLSHeaderValue, "value for TLS connection header")
+	f.StringVar(&cfg.Proxy.RequestID, "proxy.header.requestid", defaultConfig.Proxy.RequestID, "header for reqest id")
 	f.StringVar(&gzipContentTypesValue, "proxy.gzip.contenttype", defaultValues.GZIPContentTypesValue, "regexp of content types to compress")
 	f.StringSliceVar(&listenerValue, "proxy.addr", defaultValues.ListenerValue, "listener config")
 	f.KVSliceVar(&certSourcesValue, "proxy.cs", defaultValues.CertSourcesValue, "certificate sources")
--- a/config/load_test.go
+++ b/config/load_test.go
@@ -303,6 +303,13 @@ func TestLoad(t *testing.T) {
 			},
 		},
 		{
+			args: []string{"-proxy.header.requestid", "value"},
+			cfg: func(cfg *Config) *Config {
+				cfg.Proxy.RequestID = "value"
+				return cfg
+			},
+		},
+		{
 			args: []string{"-proxy.gzip.contenttype", `^text/.*$`},
 			cfg: func(cfg *Config) *Config {
 				cfg.Proxy.GZIPContentTypes = regexp.MustCompile(`^text/.*$`)
--- a/proxy/http_integration_test.go
+++ b/proxy/http_integration_test.go
@@ -50,6 +50,31 @@ func TestProxyProducesCorrectXffHeader(t *testing.T) {
 	}
 }
 
+func TestProxyRequestIDHeader(t *testing.T) {
+	got := "not called"
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		got = r.Header.Get("X-Request-ID")
+	}))
+	defer server.Close()
+
+	proxy := httptest.NewServer(&HTTPProxy{
+		Config:    config.Proxy{RequestID: "X-Request-Id"},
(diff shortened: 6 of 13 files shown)


proxy/http_headers_test.go
```diff
@@ -235,6 +235,19 @@ func TestAddHeaders(t *testing.T) {
...
+
+		{"set uniquerequestid header",
+			&http.Request{RemoteAddr: "1.2.3.4:5555", Host: "5.6.7.8:1234"},
+			config.Proxy{UniqueRequestID: "X-Unique-Request-Id"},
+			http.Header{
+				"Forwarded":           []string{"for=1.2.3.4; proto=http"},
+				"X-Forwarded-Proto":   []string{"http"},
+				"X-Forwarded-Port":    []string{"1234"},
+				"X-Real-Ip":           []string{"1.2.3.4"},
+				"X-Unique-Request-Id": []string{""}, // XXX nedd to figure out how to regexp match
```
magiconair: You add a `func UUID()` to the proxy which you can mock.

fabio.properties
```diff
@@ -353,6 +353,13 @@
 # proxy.header.tls.value =
 
 
+# proxy.header.uniquerequesti configures the header for the adding unique request id.
```
magiconair: typo

config/config.go
```diff
@@ -57,6 +57,7 @@ type Proxy struct {
 	TLSHeader             string
 	TLSHeaderValue        string
 	GZIPContentTypes      *regexp.Regexp
+	UniqueRequestID       string
```
magiconair: I think `RequestID` is enough since they need to be unique for this to make sense.

proxy/http_headers_test.go
bkmit: Made changes you suggest.

fabio.properties
bkmit: Fixed.

config/config.go
bkmit: Shortened option name

proxy/http_headers_test.go
bkmit: Logging can be made by using $header.ConfiguredHederName format already present in fabio.

magiconair:
you're also correct with the logging.

magiconair:
Almost there. For testing I suggest you add a `UUID` function to the proxy similar to the `Time` function:

```go
// UUID returns a unique id in uuid format.
// If UUID is nil, uuid.NewV4() is used.
UUID func() string
```

Then create in the `uuid` package:

```go
func NewV4() string {
    return EncodeHex(NewRandom())
}
```

Now you can mock this in the test.

magiconair:
One thing I'd like to know though and maybe you can add a benchmark for this is how many UUIDs this approach can generate? Is this using a pseudo random number generator which can generate enough entropy for thousands of req/sec or will this become a limiting factor for the throughput? If yes, then we need to find a better way.

magiconair:
There are users with 20k req/sec (or more) on a single instance. I won't name you in public but you know who you are :)

leprechau:
Quick results of some simple benchmarks of what appeared to be the 3 most popular UUID packages on github.


__Test__
```go
package main

import (
	"testing"

	guuid "github.com/google/uuid"
	muuid "github.com/m4rw3r/uuid"
	suuid "github.com/satori/go.uuid"
)

func BenchmarkV4Google(b *testing.B) {
	for i := 0; i < b.N; i++ {
		guuid.NewRandom()
	}
}

func BenchmarkV4Satori(b *testing.B) {
	for i := 0; i < b.N; i++ {
		suuid.NewV4()
	}
}

func BenchmarkV4m4rw3r(b *testing.B) {
	for i := 0; i < b.N; i++ {
		muuid.V4()
	}
}
```

__Result__
```
ahurt1:uuid-test ahurt$ go test -bench . -benchtime 10s
BenchmarkV4Google-4   	10000000	      1243 ns/op
BenchmarkV4Satori-4   	10000000	      1316 ns/op
BenchmarkV4m4rw3r-4   	10000000	      1318 ns/op
PASS
ok  	github.com/leprechau/uuid-test	42.637s
ahurt1:uuid-test ahurt$
```

In short, they all appear to be about the same but are all around 10,000 per second which is half of the stated desired requirement.

leprechau:
Think I found a winner ... it's not RFC 1422 compliant but it does generate unique IDs and it does it very fast.

__Code__

```go
package main

import (
	"fmt"
	"testing"

	guuid "github.com/google/uuid"
	muuid "github.com/m4rw3r/uuid"
	fuuid "github.com/rogpeppe/fastuuid"
	suuid "github.com/satori/go.uuid"
)

func BenchmarkV4Google(b *testing.B) {
	for i := 0; i < b.N; i++ {
		guuid.NewRandom()
	}
}

func BenchmarkV4Satori(b *testing.B) {
	for i := 0; i < b.N; i++ {
		suuid.NewV4()
	}
}

func BenchmarkV4m4rw3r(b *testing.B) {
	for i := 0; i < b.N; i++ {
		muuid.V4()
	}
}

func BenchmarkFastUUIDRaw(b *testing.B) {
	g := fuuid.MustNewGenerator()
	for i := 0; i < b.N; i++ {
		g.Next()
	}
}

func BenchmarkFastUUIDFormatted(b *testing.B) {
	g := fuuid.MustNewGenerator()
	for i := 0; i < b.N; i++ {
		_ = fmt.Sprintf("%x\n", g.Next())
	}
}
```

__Results__
```
ahurt1:uuid-test ahurt$ go test -bench . -benchtime 10s
BenchmarkV4Google-4            	10000000	      1292 ns/op
BenchmarkV4Satori-4            	10000000	      1272 ns/op
BenchmarkV4m4rw3r-4            	10000000	      1285 ns/op
BenchmarkFastUUIDRaw-4         	1000000000	        20.0 ns/op
BenchmarkFastUUIDFormatted-4   	30000000	       505 ns/op
PASS
ok  	github.com/leprechau/uuid-test	80.111s
ahurt1:uuid-test ahurt$
```

Even with the formatting being done by the relatively heavy `fmt` package the `fastuuid` package is the clear winner.

magiconair:
@leprechau very cool. Thx for digging. Want to add a fast formatter? That lib is also small enough to be vendored in.

leprechau:
@magiconair Done.  I put the formatter and test here: https://github.com/leprechau/uuid-test
@bkmit Feel free to use/adopt as needed.

bkmit:
Wow! Great research! I'll definitely use fastest library.

bkmit:
@magiconair Do I need to make some more improvements?

magiconair:
Sorry. Got sidetracked. Looking now.

fabio.properties
```diff
@@ -353,6 +353,13 @@
 # proxy.header.tls.value =
 
 
+# proxy.header.requestid configures the header for the adding request id.
```
magiconair: `proxy.header.requestid configures the header for the adding a unique request id.`

proxy/http_headers.go
```diff
@@ -72,6 +73,20 @@ func addHeaders(r *http.Request, cfg config.Proxy) error {
 	return nil
 }
 
+// addRequestID Set header with name from `cfg.RequestID` to UUID value
+//
+func addRequestID(r *http.Request, p *HTTPProxy) {
+	if p.Config.RequestID != "" {
```
magiconair: Can you add this to the `addHeaders()` function and move your test to the `TestAddHeaders()` fn as well, please? This isn't complex enough for a separate fn/test IMO. 

```go
if p.Config.RequestID != "" {
    id := p.UUID
    if id == nil {
        id = uuid.NewUUID
    }
    r.Header.Set(p.Config.RequestID, id())
}

bkmit: I have did it in separate function because it need HTTPProxy, which is not available in addHeaders.
Changing addHeaders signature lead to massive change of it unit tests.

magiconair: I see. Then just inline the couple lines into the HTTP proxy and use the integration test to test it.

magiconair:
Looks good. One last request. Can you split this up into just two commits? One for the vendored in library and one for the feature change? I don't want to squash those two.

magiconair:
Once you're done pls force push the change so that it replaces the current list of commits.

magiconair:
merged it.

magiconair:
@bkmit Thanks for your patience :)

bkmit:
@magiconair Many thanks for your patience! This was my first public golang development (micro) project!

magiconair:
For that the quality was quite high! Hats off.

Issue #294, opened 2017-05-12
Using upstream hostname for request


CGreenburg:
My team and I are working on figuring out a service discovery architecture for our multitude of Serverless stacks.

I've created two EC2 instances in AWS: one with Consul and the other with Fabio. I have a Serverless stack deployed with a single endpoint: /status. I registered a service in Consul with the address being the API Gateway URL (`xxxxxxx.execute-api....com`) and the port to 443. I set the tag to `urlprefix-/status proto=https`.

Now, maybe I'm confused on how all of this works, but how come I can't do `curl <fabio-ip>:9999/status`? When I run that command, I get a 403 status code and a body saying bad request from CloudFront. Adding a Trace header prints `Routing to service status on https://xxxxxxx.execute-api....com:443/status`. If I run `curl https://xxxxxxx.execute-api....com:443/status`, I get back a response from my lambda function as I expect.

It seems in order for it work, I need to add a Host header matching the API Gateway URL: `curl -H 'Host: xxxxxxx.execute-api....com' <fabio-ip>:9999/status`. That works fine and returns back a response from my lambda function, but that defeats the whole purpose. If I have clients go to curl the api, I can't expect them to know the API Gateway URL since Fabio and Consul are meant to abstract it.

So, my main question is, what am I doing wrong? Am I thinking about this in the wrong way? Am I missing some configuration of some sort? Thanks.

magiconair:
That should work. You're sure that your `urlprefix-` tag does not contain the hostname?

mitchelldavis:
To piggy back on @CGreenburg question.  We're hoping to get Fabio to work like a reverse proxy (I think).  HTTP -> HTTPS in this case so our users can simply hit Fabio and Fabio routes the traffic to the backend services.  (An obvious use case for the tool.)  However, we're stubbing our toes somewhere.

```
http://<fabio machine>/<service name>/<path> --> https://<service address>/<path>
```

# I think I understand that a straight through proxy wouldn't work in this case because the traffic would need to be terminated in Fabio and then continued on to the service and back in order to handle the certificates.  
# We were able to get this to work if we added the `Host: <service address>` to the request which is prohibitive.  If the software knew the host header then it could go strait there.

What are we missing?

mitchelldavis:
Here is my configuration for a Serverless REST API in Fabio:

```
/registrar	https://<garble>.amazonaws.com:443	proto=https tlsskipverify=true	100.00%
```

[...]

PR #301, opened 2017-05-27
Using upstream hostname for request (#294)

mitchelldavis:


mitchelldavis:
* So if the two tests are redundant, should I roll them into one or just get rid of the second?
* I set UseUpstreamHostname on line 73 in `route.go`.  Is that not correct?
* `usehost` works for me and the only one I can come up with otherwise is `resetHost`.

magiconair:
You can replace the tests with this:

```go
func TestProxyUsehost(t *testing.T) {
	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		fmt.Fprint(w, r.Host)
	}))

	proxy := httptest.NewServer(&HTTPProxy{
		Transport: &http.Transport{
			Dial: func(network, addr string) (net.Conn, error) {
				addr = server.URL[len("http://"):]
				return net.Dial(network, addr)
			},
		},
		Lookup: func(r *http.Request) *route.Target {
			routes := "route add mock /usehost http://a.com/ opts \"usehost\"\n"
			routes += "route add mock / http://a.com/"
			tbl, _ := route.NewTable(routes)
			return tbl.Lookup(r, "", route.Picker["rr"], route.Matcher["prefix"])
		},
	})
	defer proxy.Close()

	check := func(uri, host string) {
		resp, body := mustGet(proxy.URL + uri)
		if got, want := resp.StatusCode, http.StatusOK; got != want {
			t.Fatalf("got status %d want %d", got, want)
		}
		if got, want := string(body), host; got != want {
			t.Fatalf("got body %q want %q", got, want)
		}
	}

	proxyHost := proxy.URL[len("http://"):]
	t.Run("use host", func(t *testing.T) { check("/usehost", "a.com") })
	t.Run("no host", func(t *testing.T) { check("/", proxyHost) })
}
```

2. I mean to use `usehost` instead of `usehost=(true|false)`. This changes on how you set the `t.UseHost` flag since checking for an empty string is not sufficient. 

3. `usehost` it is then. Pls also use `t.UseHost` internally.

Thx a lot

route/route.go
```diff
@@ -71,7 +71,11 @@ func (r *Route) addTarget(service string, targetURL *url.URL, fixedWeight float6
 	if r.Opts != nil {
 		t.StripPath = r.Opts["strip"]
 		t.TLSSkipVerify = r.Opts["tlsskipverify"] == "true"
-		t.UseUpstreamHostname = r.Opts["useupstreamhostname"] == "true"
+		// If the 'usehost' key exists in the map, we set
```
magiconair: This is redundant. Just use what I wrote initially

```go
_, t.UseHost = r.Opts["usehost"]
```

route/target.go
```diff
@@ -21,10 +21,10 @@ type Target struct {
 	// TLS connections.
 	TLSSkipVerify bool
 
-	// UseUpstreamHostname causes the origin host header to be
+	// UseHost causes the origin host header to be
```
magiconair:
```
UseHost determines whether the the origin or the target hostname
is used for the upstream request
```

route/route.go
```diff
@@ -73,9 +73,7 @@ func (r *Route) addTarget(service string, targetURL *url.URL, fixedWeight float6
 		t.TLSSkipVerify = r.Opts["tlsskipverify"] == "true"
 		// If the 'usehost' key exists in the map, we set
```
magiconair: Can you drop the comment, too please? Don't want to be overly pick but it really is redundant.

mitchelldavis: No Worries.  It looks like I need to rerun the build anyways. :-D

magiconair:
I'll let this sit for another day or two before I merge it. `usehost` is concise but also ambiguous. I'll see whether I can come up with something better. Thx for your patience.

mitchelldavis:
@magiconair, how about `usedesthost`?

magiconair:
How about `host=dst`?

mitchelldavis:
ooooo.... that's way better.  Default will be `src`?

magiconair:
Yes. I think that makes more sense and provides some flexibility for the future. boolean flags are always somewhat limiting unless you can be absolutely sure that there will ever only be two choices.

magiconair:
Can you make that change?

mitchelldavis:
I'm on it!

magiconair:
Two small changes and then we can merge it.

mitchelldavis:
Thats a bummer.  It looks like an unstable test. :-(

Issue #274, opened 2017-04-26
Add parameter to vault token renewal


deuch:
Fabio need to have a parameter in the fabio.properties to let the user set the renewal interval for the Vault Token. It's 1h actually and it's too long :-) We've issue because our tokens have a 20mn lease.

magiconair:
Sure. Was expecting that this would happen at some time.

magiconair:
@deuch I've pushed a change that should make the token renew interval configurable. You have to set the `renewtoken` option in the `proxy.cs` statement. Could you check whether that solves the problem for you?

deuch:
Hello,

I've made the test and it doesn't seem to work, fabio doesn't want to start, telling me that the port is already used ... Stop fabio, kill every instance, check with netstat (root) and nothing is listening on that port.

Revert back to 1.4.3 and it works fine (with exact same properties file) ...

Please find the logs : 

2017/04/27 09:31:25 [INFO] Runtime config
{
    "Proxy": {
        "Strategy": "rr",
        "Matcher": "prefix",
        "NoRouteStatus": 404,
        "MaxConn": 10000,
        "ShutdownWait": 0,
        "DialTimeout": 30000000000,
        "ResponseHeaderTimeout": 0,
        "KeepAliveTimeout": 0,
        "FlushInterval": 1000000000,
        "LocalIP": "192.163.167.143",
        "ClientIPHeader": "",
        "TLSHeader": "",
        "TLSHeaderValue": "",
        "GZIPContentTypes": null
    },
    "Registry": {
        "Backend": "consul",
        "Static": {
            "Routes": ""
        },
        "File": {
            "Path": ""
        },
        "Consul": {
            "Addr": "dckdevdck075.mydomain.com:8543",
            "Scheme": "https",
            "Token": "xxxxxxxxxxxxxxxxxxxxxxxxx",
            "KVPath": "/l7demo1/config",
            "TagPrefix": "urlprefix-",
            "Register": true,
            "ServiceAddr": ":9998",
            "ServiceName": "l7demo1",
            "ServiceTags": null,
            "ServiceStatus": [
                "passing"
            ],
            "CheckInterval": 1000000000,
            "CheckTimeout": 3000000000,
            "CheckScheme": "http",
            "CheckTLSSkipVerify": false
        },
        "Timeout": 10000000000,
        "Retry": 500000000
    },
    "Listen": [
        {
            "Addr": "dckdevdck075:11443",
            "Proto": "tcp+sni",
[...]

PR #314, opened 2017-06-28
Issue #274: Avoid premature vault token renewals

pschultz:
Don't attempt Vault token renewals if the token isn't renewable. If it
is, don't renew the token with each refresh; only do so if the token
would expire shortly. Increase the token's lifetime by its original TTL.

This now requires the "read" capability for the "auth/token/lookup-self" path in Vault. That is granted by default to all tokens. If the lookup-self request fails for any reason, fabio will assume that the token is not renewable.

No new configuration is required. Automatic renewal will work as long as the refresh interval is smaller than the token TTL, same as before.

### DIFF
--- a/cert/vault_source.go
+++ b/cert/vault_source.go
@@ -228,7 +228,7 @@ func (s *VaultSource) renewToken(c *api.Client) error {
 	case ttl < 2*s.Refresh:
 		// Renew the token if it isn't valid for two more refresh intervals.
 		break
-	case ttl < 1*time.Minute:
+	case ttl < time.Minute:
 		// Renew the token if it isn't valid for one more minute. This happens
 		// if s.Refresh is small, say one second. It is risky to renew the
 		// token just one or two seconds before expiration; networks are
--- a/cert/vault_source.go
+++ b/cert/vault_source.go
@@ -32,11 +32,17 @@ type VaultSource struct {
 	mu         sync.Mutex
 	vaultToken string // VAULT_TOKEN env var. Might be wrapped.
 	auth       struct {
-		token      string    // actual token
-		expireTime time.Time // zero value if the token isn't renewable
+		// token is the actual Vault token.
+		token string
 
-		// The desired token lifetime after renewal, in seconds. This value is
-		// advisory and the Vault server may ignore or silently change it.
+		// expireTime is the time at which the token expires (becomes useless).
+		// The zero value indicates that the token is not renewable or never
+		// expires.
+		expireTime time.Time
+
+		// renewTTL is the desired token lifetime after renewal, in seconds.
+		// This value is advisory and the Vault server may ignore or silently
+		// change it.
 		renewTTL int
 	}
 }
@@ -133,12 +139,13 @@ func (s *VaultSource) checkRenewal(c *api.Client) {
 		return
 	case dropNotRenewableWarning:
 		return
+	default:
+		ttl := time.Until(data.ExpireTime)
+		ttl = ttl / time.Second * time.Second // truncate to seconds
+		log.Printf("[WARN] vault: Token is not renewable and will expire %s from now at %s",
+			ttl, data.ExpireTime.Format(time.RFC3339))
 	}
 
-	ttl := time.Until(data.ExpireTime)
-	ttl = ttl / time.Second * time.Second // truncate to seconds
-	log.Printf("[WARN] vault: Token is not renewable and will expire %s from now (at %s)",
-		ttl, data.ExpireTime.Format(time.RFC3339))
 }
 
 func (s *VaultSource) LoadClientCAs() (*x509.CertPool, error) {
--- a/cert/vault_source.go
+++ b/cert/vault_source.go
@@ -44,8 +44,6 @@ type VaultSource struct {
 		// This value is advisory and the Vault server may ignore or silently
 		// change it.
 		renewTTL int
-
-		once sync.Once
 	}
 }
 
@@ -71,13 +69,10 @@ func (s *VaultSource) client() (*api.Client, error) {
 
 func (s *VaultSource) setAuth(c *api.Client) error {
 	s.mu.Lock()
-	defer func() {
-		c.SetToken(s.auth.token)
-		s.auth.once.Do(func() { s.checkRenewal(c) })
-		s.mu.Unlock()
-	}()
+	defer s.mu.Unlock()
 
 	if s.auth.token != "" {
+		c.SetToken(s.auth.token)
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


pschultz:
The builds failed because of network issues between Travis and HashiCorp:

> $ wget https://releases.hashicorp.com/consul/0.8.3/consul_0.8.3_linux_amd64.zip
> --2017-06-28 14:01:50--  https://releases.hashicorp.com/consul/0.8.3/consul_0.8.3_linux_amd64.zip
> Resolving releases.hashicorp.com (releases.hashicorp.com)... 151.101.89.183, 2a04:4e42:11::439
> Connecting to releases.hashicorp.com (releases.hashicorp.com)|151.101.89.183|:443... failed: Connection timed out.
> Connecting to releases.hashicorp.com (releases.hashicorp.com)|2a04:4e42:11::439|:443... failed: Network is unreachable.

I don't think I can retry them.

cert/vault_source.go
```diff
@@ -45,20 +50,27 @@ func (s *VaultSource) client() (*api.Client, error) {
...
+	if err := s.setAuth(c); err != nil {
 		return nil, err
 	}
+
 	return c, nil
 }
 
-func (s *VaultSource) setToken(c *api.Client) error {
+func (s *VaultSource) setAuth(c *api.Client) error {
+	firstCall := true
```
magiconair: This looks racy and should probably be inside the lock.

cert/vault_source.go
```diff
@@ -71,17 +83,53 @@ func (s *VaultSource) setToken(c *api.Client) error {
...
 
+// dropNotRenewableWarning controls whether the 'Token is not renewable'
+// warning is logged. This is useful for testing where this is the expected
+// behavior. On production, this should always be set to false.
+var dropNotRenewableWarning bool
+
+// checkRenewal checks if the Vault token can be renewed, and if so when it
+// expires and how big the renewal increment should be.
+func (s *VaultSource) checkRenewal(c *api.Client) {
+	response, err := c.Auth().Token().LookupSelf()
```
magiconair: `s/response/resp/`

cert/vault_source.go
```diff
@@ -71,17 +83,53 @@ func (s *VaultSource) setToken(c *api.Client) error {
...
+		CreationTTL int       `json:"creation_ttl"`
+		ExpireTime  time.Time `json:"expire_time"`
+		Renewable   bool      `json:"renewable"`
+	}
+	if err := json.Unmarshal(b, &data); err != nil {
+		log.Printf("[WARN] vault: lookup-self failed, token renewal is disabled: %s", err)
+		return
+	}
+
+	if data.Renewable {
```
magiconair: pls rewrite as `switch` statement. I try to avoid if/else cascades.

cert/vault_source.go
```diff
@@ -165,3 +201,37 @@ func (s *VaultSource) load(path string) (pemBlocks map[string][]byte, err error)
...
+		// if s.Refresh is small, say one second. It is risky to renew the
+		// token just one or two seconds before expiration; networks are
+		// unreliable, clocks can be skewed, etc.
+		break
+	default:
+		// Token doesn't need to be renewed yet.
+		return nil
+	}
+
+	response, err := c.Auth().Token().RenewSelf(s.auth.renewIncrement)
```
magiconair: `s/response/resp/`

cert/vault_source.go
```diff
@@ -165,3 +201,37 @@ func (s *VaultSource) load(path string) (pemBlocks map[string][]byte, err error)
...
+
+	ttl := time.Until(s.auth.expireTime)
+	switch {
+	case s.auth.expireTime.IsZero():
+		// Token isn't renewable.
+		return nil
+	case ttl < 2*s.Refresh:
+		// Renew the token if it isn't valid for two more refresh intervals.
+		break
+	case ttl < 1*time.Minute:
```
magiconair: pls drop the `1*` since it is redundant.

cert/vault_source.go
```diff
@@ -29,8 +30,12 @@ type VaultSource struct {
 	Refresh      time.Duration
 
 	mu         sync.Mutex
-	token      string // actual token
 	vaultToken string // VAULT_TOKEN env var. Might be wrapped.
+	auth       struct {
+		token          string    // actual token
+		expireTime     time.Time // zero value if the token isn't renewable
+		renewIncrement int
```
magiconair: pls add a comment and explain the unit. Maybe a shorter name, e.g. `renewTTL`

cert/vault_source.go
pschultz: It is redundant, but it reads as "one minute" which I prefer. This is a constant expression that will be evaluated at compile time, so there is no runtime impact. If you insist I'll change it though.

cert/vault_source.go
pschultz: Technically not a race (all reads and writes are inside the lock), but I moved it down anyway to remove any doubts.

pschultz:
Please take another look. I will fixup into a single commit once you're good with the PR.

cert/vault_source.go
```diff
@@ -32,9 +32,12 @@ type VaultSource struct {
 	mu         sync.Mutex
 	vaultToken string // VAULT_TOKEN env var. Might be wrapped.
 	auth       struct {
-		token          string    // actual token
-		expireTime     time.Time // zero value if the token isn't renewable
-		renewIncrement int
+		token      string    // actual token
```
magiconair: pls use the Go doc standard here:

```
// token is the actual token
token string

// expireTime is ...
expireTime time.Time
...
```

cert/vault_source.go
```diff
@@ -119,15 +124,21 @@ func (s *VaultSource) checkRenewal(c *api.Client) {
...
+	case data.ExpireTime.IsZero():
+		// token doesn't expire
+		return
+	case dropNotRenewableWarning:
+		return
 	}
+
+	ttl := time.Until(data.ExpireTime)
+	ttl = ttl / time.Second * time.Second // truncate to seconds
+	log.Printf("[WARN] vault: Token is not renewable and will expire %s from now (at %s)",
```
magiconair: The logging is the `default` case, right?

pls change the log msg to `Token is not renewable and expires in %s from now at %s`

magiconair:
Don't bother. I can squash the merge myself.

cert/vault_source.go
magiconair: I don't insist but I'd prefer since it is redundant :)

cert/vault_source.go
magiconair: Right, since this is a local var but then this doesn't make sense. The `defer` function should then read:

```go
defer func() {
    c.SetToken(s.auth.token)
    if s.auth.token != "" {
        s.checkRenewal(c)
    }
    s.mu.Unlock()
}
```

magiconair:
@pschultz Cheeky side question: Would you mind me adding your logo to the README? :)

cert/vault_source.go
pschultz: s.auth.token will never be empty when the deferred function executes (assuming there is no error). Apparently I made this is too confusing. I think we can use sync.Once to make it obvious. Let my try that.

magiconair: I also suggest to remove the `SetToken` call from the `defer` method and put it where necessary. There is too much magic going on here which is hard to follow.

pschultz: See 49767dd

pschultz:
Sure, you can add the logo. We don't have an English site you can link to though, only the one in German.

magiconair:
I think this is more readable. What do you think?

```go
func (s *VaultSource) setAuth(c *api.Client) error {
	s.mu.Lock()
	defer s.mu.Unlock()

	if s.auth.token != "" {
		return nil
	}

	if s.vaultToken == "" {
		return errors.New("vault: no token")
	}

	// did we get a wrapped token?
	resp, err := c.Logical().Unwrap(s.vaultToken)
	if err != nil {
		// unwrapping failed?
		if !strings..HasPrefix(err.Error(), "no value found at") {
			return err
		}
		s.auth.token = s.vaultToken
	} else {
		log.Printf("[INFO] vault: Unwrapped token %s", s.vaultToken)
		s.auth.token = resp.Auth.ClientToken
	}

	s.auth.once.Do(func() { s.checkRenewal(c) })
	c.SetToken(s.auth.token)
	return nil
}
```

pschultz:
Yeah, that's better. The last if/else statements can even be done with a nice switch. See f6e2463.

magiconair:
LGTM

Issue #135, opened 2016-07-26
Auto-generated Vault certs


far-blue:
I'm investigating hooking up Vault with Fabio and I'm a little confused about the cert store config. The documentation suggests Fabio will fetch certs dynamically from Vault but the config seems to imply that the certs are expected to be manually dropped into a generic 'secret' store in Vault. Is there also support for the PKI secret store and its ability to auto-generate certificates?

magiconair:
Not yet. I'll check what it takes to add that.

far-blue:
I can manually issue a cert using the following:

> vault write pki/issue/server-dev common_name=foo.server.dev

In this example I've set up a pki backend with an internally generated root cert for *.server.dev and then I've created a role 'server-dev' that is allowed to serve certs for <something>.server.dev based on the root cert. Using the above command will return both the private and public parts of the new cert as well as a copy of the CA.

blalor:
I'd really like to see Vault's PKI backend supported, as well.   I've been chewing on this for an hour or so and I think the model of multiple certs per cert store doesn't exactly mesh with Vault's PKI backend, but if each `vault-pki` certificate source has a single pki secret path instead of multiples, that would probably work fine. One of the config options to the `vault-pki` cert source would be a list of SANs that would be on the cert, perhaps?

magiconair:
I'm curious how this would work independent of vault since you could mimic this with a script or LetsEncrypt or something else. The current CertSource model is that certificates exist and that fabio just picks them up. If a cert does not exist fabio will refuse the connection. 

In the model of a generated cert that would no longer happen unless the cert provider refuses to generate the cert.

Who controls which certs are permissible? Can fabio just generate any cert? What if I announce `urlprefix-www.google.com/` ?

pschultz:
The Vault operator defines what certificates can be issued on a given path.

From fabios perspective there is almost no difference to fetching certs from the generic secret backend. In Go, it looks something like this:

```go
import "github.com/hashicorp/vault/api"

var client *api.Client

response, err := client.Logical().Write("pki-fabio/issue/example.com", map[string]interface{}{
	"common_name": "foo.example.com",
	"alt_names": "localhost,127.0.0.1",
	"ttl":         "24h", // uses Vault's default if omitted
})

pemKey := response.Data["private_key"].(string)
certChain := response.Data["certificate"].(string) + "\n"
for _, cert := range response.Data["ca_chain"].([]interface{}) {
	certChain += cert.(string) + "\n"
[...]

PR #315, opened 2017-07-01
Vault PKI cert source

pschultz:
POC for the Vault PKI source. I have not tested this thoroughly yet, especially under load. I want to get feedback on the design first.

I moved the Vault client bits out of the existing vault source, so they can be shared with the new PKI source. I had to redo the token renewal, since that was triggered as part of the refresh logic. However, the PKI source will for the most part just sit and wait for certificates to expire. So I created a wrapper around the *api.Client, that will renew tokens after half their lifetime. This is the same behaviour as in Nomad and Consul Template, for instance.

I left a few TODO comments in the code. I'd appreciate your thoughts on those.

### DIFF
--- a/vendor/vendor.json
+++ b/vendor/vendor.json
@@ -2,45 +2,252 @@
 	"comment": "",
 	"ignore": "test",
 	"package": [
-		{"path":"github.com/armon/go-proxyproto","checksumSHA1":"eAall4ACaMG40mzSJ5Oc95GiF1A=","revision":"609d6338d3a76ec26ac3fe7045a164d9a58436e7","revisionTime":"2015-02-06T18:58:55-08:00"},
-		{"path":"github.com/circonus-labs/circonus-gometrics","checksumSHA1":"ZAdLZ0e/hin4AXxdS9F8y0yi/bg=","revision":"f8c68ed96a065c10344dcaf802f608781fc0a981","revisionTime":"2016-08-30T16:47:25Z"},
-		{"path":"github.com/circonus-labs/circonus-gometrics/api","checksumSHA1":"2hV0W0wM75u+OJ4kGv0i7B95cmY=","revision":"f8c68ed96a065c10344dcaf802f608781fc0a981","revisionTime":"2016-08-30T16:47:25Z"},
-		{"path":"github.com/circonus-labs/circonus-gometrics/checkmgr","checksumSHA1":"r3XLF0s89Y7UJgS9YP3NbdyafNY=","revision":"f8c68ed96a065c10344dcaf802f608781fc0a981","revisionTime":"2016-08-30T16:47:25Z"},
-		{"path":"github.com/circonus-labs/circonusllhist","checksumSHA1":"C4Z7+l5GOpOCW5DcvNYzheGvQRE=","revision":"d724266ae5270ae8b87a5d2e8081f04e307c3c18","revisionTime":"2016-05-26T04:38:13Z"},
-		{"path":"github.com/cyberdelia/go-metrics-graphite","checksumSHA1":"8/Q1JbAHUmL4sDURLq6yron4K/I=","revision":"b8345b7f01d571b05366d5791a034d872f1bb36f","revisionTime":"2015-08-25T20:22:00-07:00"},
-		{"path":"github.com/fatih/structs","checksumSHA1":"zKLQNNEpgfrJOwVrDDyBMYEIx/w=","revision":"5ada2f449b108d87dbd8c1e60c32cdd065c27886","revisionTime":"2016-06-01T09:31:17Z"},
-		{"path":"github.com/hashicorp/consul/api","checksumSHA1":"GsJ84gKbQno8KbojhVTgSVWNues=","revision":"402636ff2db998edef392ac6d59210d2170b3ebf","revisionTime":"2017-04-05T04:22:14Z","version":"v0.8.0","versionExact":"v0.8.0"},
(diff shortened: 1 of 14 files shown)


magiconair:
What I understand is that you are doing two things here:

1. refactor the existing code to allow for dynamic cert generation
2. add vault pki as a dynamic cert generator

What I usually do in these cases is to split the change into two commits: first the functional equivalent refactoring, then add the new feature on top of that new functionality.

I like the `Issuer` interface but the first thing that came to mind is whether the cert store setup could be extended in general with different kind of issuers. Static onces, dynamic ones, ... That was the idea of the `Store` which would become a `Cache`. The other thing I was thinking of when I read this was what about LetsEncrypt? 

Please note that this isn't fully thought out but it might make sense to look at the architecture of the cert stores as a whole and not just for Vault.

As a benefit I've learned about `golang.org/x/sync/singleflight` :)

One other thing caught my eye: the setting of `StrictMatch` in `main.go` if the source is an issuer. I think this should be caught during config loading and tested there and this should trigger an INFO message in the log.

pschultz:
I will re-arrange the commits to keep the Vault stuff separate, and at the same time move the StrictMatch check into the config package. While it is nice that this check can now be tested, it also means that it will have to be done for all future Issuer implementations (which is easily forgotten).

The store is unaware of any sources (and, in fact, the sources are unaware of the store). The only part that is aware of both is cert.TLSConfig. In my opionion that's a good thing.

Adding other on-demand sources doesn't require more changes to either the store or cert.TLSConfig, just a new Issuer implementation and config support. I threw together two more examples in [this gist](https://gist.github.com/pschultz/80a13aa3488f4e9afa1e440bea3c3d6e) - please handle with care; it compiles if you drop it in certs, but I haven't checked it for correctness or anything. One is for the ACME protocol (e.g. Letsencrypt) and the other takes a CA cert file and private key and issues certs directly.

The ACMESource in particular would require lots of configuration, and each challenge provider you'd want to support comes with its own rabbit hole, but none of that relates to the store I think.

By the way, as far as I know Letsencrypt will only allow one non-revoked certificate per common name at any given time, so it has to be revoked before a new certificate can be issued. If more than one fabio instance serve a domain they have to coordinate accordingly. That's out of scope of this MR though I think.

It's apparent to me now that a Store.ReplaceCertificate method would be useful. Then the Issuers don't have to keep track of all the certs they ever issued. Apart from that I don't see any beneficial changes I could make to the store.

magiconair:
@pschultz This has fallen a bit through the cracks for various reasons. I'd like to pick this one up again and merge it if you still feel comfortable about this.  Could you rebase this, please?

pschultz:
Rebased on master.

vendor/vendor.json
```diff
@@ -2,45 +2,252 @@
...
+		{
+			"checksumSHA1": "xvuVB+aMh4udpinGsmtK31pcWC0=",
+			"path": "golang.org/x/net/websocket",
+			"revision": "9946ad7d5eae91d8edca4f54d1a1e130a052e823",
+			"revisionTime": "2015-10-20T22:50:55Z"
+		},
+		{
+			"checksumSHA1": "VhUZFUuhLFSBFUfskMC4am5RIdc=",
+			"path": "golang.org/x/sync/singleflight",
+			"revision": "8e0aa688b654ef28caa72506fa5ec8dba9fc7690",
```
pschultz: The very last entry in this list is new (singleflight), everything else only re-formatted. Did you change the vendor tool?

magiconair: I've written a tool https://github.com/magiconair/vendorfmt to reformat the JSON to a more merge-friendly single-line format. If you run `make vendorfmt` it should format it.

magiconair:
Lets merge this.

magiconair:
Fixes #135

magiconair:
Thanks a lot @pschultz !

magiconair:
@pschultz The VaultPKI integration fails with go1.10rc1. I'm getting this. Would you be OK having a look at this?

```
$ go test -run TestVaultPKISource ./cert
--- FAIL: TestVaultPKISource (2.49s)
	source_test.go:305: Starting vault: "\x1b[0;0mVault v0.9.3 ('5acd6a21d5a69ab49d0f7c0bf540123a9b2c696d')\x1b[0m\n"
    --- FAIL: TestVaultPKISource/renewable_token (0.46s)
    	source_test.go:564: got Get https://127.0.0.1:49363: x509: certificate signed by unknown authority want nil
    --- FAIL: TestVaultPKISource/non-renewable_token (0.26s)
    	source_test.go:564: got Get https://127.0.0.1:49370: x509: certificate signed by unknown authority want nil
    --- FAIL: TestVaultPKISource/renewable_orphan_token (0.30s)
    	source_test.go:564: got Get https://127.0.0.1:49378: x509: certificate signed by unknown authority want nil
    --- FAIL: TestVaultPKISource/non-renewable_orphan_token (0.36s)
    	source_test.go:564: got Get https://127.0.0.1:49390: x509: certificate signed by unknown authority want nil
    --- FAIL: TestVaultPKISource/renewable_wrapped_token (0.34s)
    	source_test.go:564: got Get https://127.0.0.1:49400: x509: certificate signed by unknown authority want nil
    --- FAIL: TestVaultPKISource/non-renewable_wrapped_token (0.29s)
    	source_test.go:564: got Get https://127.0.0.1:49413: x509: certificate signed by unknown authority want nil
FAIL
FAIL	github.com/fabiolb/fabio/cert	2.507s
```

pschultz:
Will do, but not before the 12th because I'm off on vacation.

magiconair:
@pschultz I've found it. See #434

PR #321, opened 2017-07-24
Fix memory leak in tcp proxy

Crypto89:
This patch wraps the ServeTCP to clean up old connections from the `s.conns` map


magiconair:
I really missed this? Wow. Thx. I'll have a closer look a bit later.

PR #328, opened 2017-08-03
Resolves #327 PoC raw_statsd

### LINKED ISSUE #327: Statsd output is not good
Aside from the actual broken protocol packets in #207 the implementation of instrumentation with statsd in this project is just really bad. Exporting gauges of pre-aggregated metrics is a serious anti-pattern to typical/normal statsd instrumentation best practices. I'm curious if there is some specific reason why this model was adopted. Is there any appetite for re-instrumenting in a more standard manner?

drawks:
This is purely a proof of concept that demonstrates what is being discussed in #327 client code is cribbed directly from github.com/etsy/statsd/examples/go and could be further reduced and cleaned up.

* adds new metric sink `raw_statsd` uses same config as `statsd`
* new sink sends singl raw events directly to statsd server without any
  preaggreation or rollup.
* TODO - use more featureful/performant statsd client implementation such as
  https://github.com/alexcesaro/statsd or https://github.com/quipo/statsd
  which implement features such as optional buffering.

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -66,6 +66,10 @@ func NewRegistry(cfg config.Metrics) (r Registry, err error) {
 		log.Printf("[INFO] Sending metrics to StatsD on %s as %q", cfg.StatsDAddr, prefix)
 		return gmStatsDRegistry(prefix, cfg.StatsDAddr, cfg.Interval)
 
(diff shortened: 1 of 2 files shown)


drawks:
Closing this in favor of work being done in #329

Issue #211, opened 2017-01-10
Support for prometheus


Equilibri0:
Hi there,

it would be great to get the fabio metrics also to a prometheus instance. 
Is there any chance for ?

greetz

Equilibri0

magiconair:
Somehow I knew that this was going to come :) I'm working on another metrics ticket right now and thought the same thing. I'll add it to the list.

raben2:
+1

pipozzz:
+1

mdirkse:
+1

deuch:
+1 for me too !

agonzalezandino:
+1 that'd be cool!

nanoz:
Any news on the Prometheus /metrics endpoint ?

This feature would be very helpful to us to measure things like requests throughput, call latencies and status codes, globally and maybe per entrypoint/endpoint as well ? :)

csawyerYumaed:
I'd love to see prom support, but in the meantime you should be able to work-around it with statsd in fabio and https://github.com/prometheus/statsd_exporter

magiconair:
I need to wrap my head around on how to configure the naming and to expose the tags. The issue is that some metrics system want a single string as name which flattens all parts of the name into a single value. Others have a base name and other parts as a tag. What makes this a bit more complex is that some of the names are auto-generated and the pattern is configurable. So either this mechanism needs to work for all providers or I need to come up with something else. It probably isn't that difficult but that's the hold-up. Same for #165 .

drawks:
Any progress on updating the internal Registry api to support labels? It would be a boon to both prometheus support as well as datadog style statsd+labels and others.

IMHO, part of the solution could be as simple as passing into the Registry's update functions a context object which contains the same information that is currently being used to flatten out into a name. And leave it up to the Registry to do the work of either flattening in accordance to the configured template OR using the attributes of the context in whatever other manner makes sense.

tbouvet:
+1

magiconair:
work on consul 1.0 has kept me quite busy. I plan to go through open issues soon.

nuriel77:
+1

magiconair:
working on it.

drawks:
Any updates here?

magiconair:
I've started and got distracted with family and personal stuff. Sorry man. I'll give it another push.

marcosnils:
[...]

PR #329, opened 2017-08-06
Add raw statsd reporter

### LINKED ISSUE #327: Statsd output is not good

magiconair:
This patch adds a statsd reporter without aggregation by using the https://github.com/alexcesaro/statsd library.

Fixes #327

### DIFF
--- a/metrics/statsd_raw.go
+++ b/metrics/statsd_raw.go
@@ -1,0 +1,63 @@
+package metrics
+
+import (
+	"errors"
+	"fmt"
+	"time"
+
+	alstatsd "github.com/alexcesaro/statsd"
+)
+
+func newRawStatsDRegistry(prefix, addr string, interval time.Duration) (Registry, error) {
+	if addr == "" {
+		return nil, errors.New(" statsd addr missing")
+	}
+
+	c, err := alstatsd.New(alstatsd.Address(addr), alstatsd.FlushPeriod(interval))
+	if err != nil {
+		return nil, fmt.Errorf(" cannot init statsd client: %s", err)
+	}
+
+	return &rawStatsDRegistry{c}, nil
+}
+
+type rawStatsDRegistry struct {
+	c *alstatsd.Client
+}
+
+func (r *rawStatsDRegistry) Names() []string        { return nil }
+func (r *rawStatsDRegistry) Unregister(name string) {}
+func (r *rawStatsDRegistry) UnregisterAll()         {}
+
+func (r *rawStatsDRegistry) GetCounter(name string) Counter {
+	return &rawStatsDCounter{r.c, name}
+}
+
+func (r *rawStatsDRegistry) GetTimer(name string) Timer {
+	return &rawStatsDTimer{r.c, name}
+}
+
+type rawStatsDCounter struct {
+	c    *alstatsd.Client
+	name string
+}
+
+func (c *rawStatsDCounter) Inc(n int64) {
+	c.c.Increment(c.name)
+}
+
+type rawStatsDTimer struct {
+	c    *alstatsd.Client
+	name string
+}
+
+func (t *rawStatsDTimer) Update(d time.Duration) {
+	t.c.Timing(t.name, int(d/time.Millisecond))
+}
+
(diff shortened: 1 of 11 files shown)


magiconair:
@drawks I've addressed your comments but I'm unsure whether I should do `prefix.name` or `prefix--name` as in the other statsd impl. The [spec](https://github.com/b/statsd_spec} isn't clear on names and prefixes. Is `--` a convention to separate prefixes from names?

drawks:
the `--` is not used in any statsd implementation I've touched. There is no notion of "prefixes" really in statsd. Some statsd client implementations allow for prepending an arbitrary string to help with namespacing metrics, but `--` is another strange quirk that seems specific to fabio...

magiconair:
This got pulled in via the statsd lib so I'm happy to change it. Maybe it is best not to add a separator at all then you can choose what you want.

drawks:
I'd just use the Prefix convenience function in the library you've chosen 

https://godoc.org/gopkg.in/alexcesaro/statsd.v2#Prefix

It looks like it uses a "." separator, so you'd have to special case when prefix is `""` to not pass the empty string.

drawks:
I've just issue a PR against your branch `issue-327-raw-statsd` to use the statsd client library's own prefixing routine.

I've done some testing with your branch and it seems to work and address the issues identified in #327 

The only lingering doubt I have about this branch is that the library you've chosen doesn't seem to have much active development and has some IMHO questionable UDP socket handling behavior (I.E. failing to initialize a statsd client if the destination isn't listening but subsequently handling re-establishing connection if there was a listener on startup. In my experience it is often the expectation that instrumentation shouldn't block a service from starting as it can often times be victimized by deployment/startup/name resolution changes and races.)

magiconair:
I'll integrated #331 but I want to take the opportunity to generally refactor the metrics approach to make supporting additional backends much simpler. I'll get rid of the `Timer` and `Counter` shims and just provide functions for metrics events just like statsd and circonus. However, I need to sweep through the code to replace metrics with just names. So bear with me for a couple more days until this is done.

I'd also like to finally attack the tags issue for DataDog and Prometheus.

drawks:
If you are actively working on adding tag support I'll pull back from my effort to do the same. I will reiterate that I /think/ the appropriate approach here is to make available all the values normally used for filling in the `metrics.name` template available in a struct stored in Route.targets and grab any other interesting client connection scoped data from the request and pass that as a `context.Context` into the `Counter`/`Timer` update functions. The template from `metrics.name` can then be passed into the `metrics.Registry` and leave it up to each metric backend to do what it wants with the 3 pieces of info.

This would allow:
* existing flat namespaced non-label metrics to operate the same
* hybrid flat+labels (DataDog) to do a combination of a templated flat name and additional labels
* name + label only (prometheus/influx/opentsdb) to have simple metric names with full (or configurable) label sets

magiconair:
I've created #334 which is the metrics refactor code and then #335 for the raw statsd code on top of that. That looks simpler but needs testing.

I haven't started with tag support and if you have an approach then I'm more than happy to use that. Can you show a proof-of-concept code so that we have a basis for discussion?

pvandervelde:
@magiconair What are the odds these statsd metrics changes are going to make it into a release in the near future? We would love to have the improved statsd metrics.

magiconair:
I’ll work on it and try to get it in the next release

magiconair:

metrics/statsd_raw.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+type rawStatsDCounter struct {
+	c    *alstatsd.Client
+	name string
+}
+
+func (c *rawStatsDCounter) Inc(n int64) {
+	c.c.Increment(c.name)
+}
```
parasyte: The input `n` arg is unused.

PR #334, opened 2017-08-11
metrics: refactor metrics to deal only with events

magiconair:
This patch refactors the metrics package to make integration with
aggregation and event based metrics libraries simpler. It drops the
Counter and Timer shim objects and exposes only functions for triggering
metric events.

Namespacing is supported but only for go-metrics based libraries.

The api no longer exposes the current rate and 99% percentile.


magiconair:

PR #335, opened 2017-08-11
Issue 327 raw statsd with refactor metrics

magiconair:


metrics/statsdraw/registry.go
```diff
@@ -0,0 +1,43 @@
...
+type registry struct {
+	c *alstatsd.Client
+}
+
+func (r *registry) Names(string) []string     { return nil }
+func (r *registry) Unregister(string, string) {}
+func (r *registry) UnregisterAll(string)      {}
+
+func (r *registry) Gauge(_, name string, n float64) {
+	r.c.Count(name, n)
```
drawks: Conflating a gauge and a counter here is going to cause problems. Counters reset between flushes unlike gauges ([at the statsd server](https://github.com/etsy/statsd/blob/master/docs/metric_types.md#gauges).)  It seems like the only time you use this is in `newRawProxy` to track the number of active conns, in order to do this correctly with a gauge you should track the actual number and report that number as a gauge.

However if you want to use the relative updates for a gauges (a questionable feature in statsd) you should always make sure that you reset the gauge value to 0 when you first update it in your app, otherwise a crash will leave you with an invalid base value.

magiconair:

PR #366, opened 2017-10-07
log: add leveled logging

magiconair:
This patch implements a simple leveled logging output to provide the
ability to have DEBUG logs which are not logged by default. The
implementation allows changing the log level at runtime but in this
implementation there is no mechanism to change the value at runtime.


magiconair:
Still need to add a test for the LevelWriter

PR #381, opened 2017-11-07
Remove unused variable

antham:


magiconair:
I'm not sure whether this enables an unwanted compiler optimization. Are the benchmark results different after you've removed the variable?

antham:
Before the change, I have this : 

```
go test -bench=BenchmarkProxyLogger github.com/fabiolb/fabio/proxy
2017/11/07 13:00:19 [INFO] cert: Store has certificates for ["example.com"]
goos: darwin
goarch: amd64
pkg: github.com/fabiolb/fabio/proxy
BenchmarkProxyLogger-4            100000             11171 ns/op
PASS
ok      github.com/fabiolb/fabio/proxy  1.781s
```

and after the change that : 

```
2017/11/07 13:01:32 [INFO] cert: Store has certificates for ["example.com"]
goos: darwin
goarch: amd64
pkg: github.com/fabiolb/fabio/proxy
BenchmarkProxyLogger-4            100000             11111 ns/op
PASS
ok      github.com/fabiolb/fabio/proxy  1.670s
```

antham:
It's the same here with http_headers_test you have from tests : 
```
proxy/http_headers_test.go:390:6: s declared but not used
```

it's for the same reason you told before this statement exists ?

magiconair:
The intention was to not let the compiler optimize a function away where the return value is never used. I may have gotten this wrong but that was the idea.

See last paragraphs in https://dave.cheney.net/2013/06/30/how-to-write-benchmarks-in-go

magiconair:
I think these two are unrelated. The var in the `BenchmarkProxyLogger` can go.

magiconair:
Thanks!

antham:
After reading this article, it's written you need to store the result in a package variable : 
```
always store the result to a package level variable
so the compiler cannot eliminate the Benchmark itself.
```

antham:
And the var in http_headers_test is necessary ? If it's the case, if think a little comment could avoid another discussion.

magiconair:
I need to check again but I agree that there should be a comment if it is necessary indeed.

Issue #87, opened 2016-04-19
Domain or protocol redirection


mazhack:
Hi again, 
i have some problems with fabio, i need redirect http traffic to https, for example
http://domain.com -> https://www.domain.com
http://www.domain.com -> https://www.domain.com

fabio does not support redirection, so i create a dockers with lighttpd and configure it to make the redirections, but i can not add route only for http or https

so if i register docker with tag:
urlprefix-domain.com/  -> one docker, redirects http and https to https://www.domain.com
urlprefix-www.domain.com/  -> register 2 dockers, one for redirect and another for application, redirects http to https, and https redirect again to https, so if i use round robin balancing, half https connections make a redirect again... 

it is possible define a protocol for route?, so route http traffic to a group of docker and https to another group of docker, for example:

urlprefix-www.domain.com:http/ -> send to redirect dockers
urlprefix-www.domain.com:https/ -> send to application dockers
urlprefix-www.domain.com:ws/ -> send to websocket dockers
urlprefix-www.domain.com:wss/ -> send to secure websocket dockers

urlprefix-/ -> send http and https, already defined
urlprefix-:http/ -> send only http, new
urlprefix-:https/ -> send only https, new
....

path, domain and protocol are optional, so it is not breaking change.

This is only an idea, a proposal, you have the last word.

mazhack:
if fabio is behind of another proxy, then this has not sense, but if is directly connected to internet is very useful

magiconair:
The way you should do this is to do the redirect in the application and not on the LB. If your application needs a resource to be on HTTPS then it should detect that and send the redirect itself. This way your application does not depend on the LB to do the work.

mazhack:
I use X-Forwarded-Proto header before, but all domain is secure, so i want to do the redirection out of application, for me is better do it in reverse proxy and not load all php application in symfony for do a redirection.
I am using now 2 fabio instances one for http and another for https and work fine. 

many thanks for the answer.

iadknet:
I agree with @mazhack and am looking for a similar solution.

I understand that service configuration should be decoupled from the load balancer, but something as basic as enforcing ssl definitely belongs with the load balancer.

magiconair:
The question is what you consider basic and universal but I agree that fabio could provide this providing that the application configures it. That means there needs to be a way for the application to express this. It's already on the list for #87.

iadknet:
I disagree with the philosophy that this is a service-level configuration.

[...]

Issue #385, opened 2017-11-09
opts with host= with multiple routes does not work as expected


discobean:
In Fabio 1.5.3 say I have 2 routes as such:

```
route add new-location-api location-api.domain.com/ http://localhost:9999 opts "host=new-location-api.domain.com"
route add old-location-api location-api.domain.com/ http://domain-location-api-production-external.dynamic.domain.com
```

The `host=xxxx` option appears on both routes.  (But notice the second route does not specify a host option)

If you put the second route above the first, then neither route has a `host=` option.

magiconair:
Found it. The opts are stored on the target but set on the route. 

https://github.com/fabiolb/fabio/blob/4dbab1b7be568100164272aa661d797cd243d006/route/table.go#L156
https://github.com/fabiolb/fabio/blob/4dbab1b7be568100164272aa661d797cd243d006/route/table.go#L162

On my way to the airport now. I’ll see what I can do before the flight. Otherwise tomorrow.

discobean:
All good mate, thanks

magiconair:
@discobean I want to write an integration test for the `host` option but if you want then you can test the patch in #388 in the meantime.

PR #395, opened 2017-11-28
Implement redirect capability

ctlajoie:
NOTE: The syntax described below was not used in the final implementation.

This is my initial attempt to implement the ability to perform redirections (see #87). I decided to indicate a redirect by prefixing the destination with the HTTP status code to be used for redirection, followed by a pipe, followed by the specific URL you want to redirect to. There is no built-in protection against creating redirect loops currently. Here is how you might use it:

```
route add svc example.com:80/ 302|https://example.com
route add svc example.com:443/ 1.2.3.4:5678
```
Note that these include explicit source ports. To use it on consul services, you have to use multiple service tags.
```
urlprefix-example.com:80/ 302|https://example.com
urlprefix-example.com:443/
```
You can create redirect routes to anywhere.
```
route add svc example.com/dothething/ 301|https://github.com/fabiolb/fabio
```

Looking for feedback.
(Signed CLA)

### DIFF
--- a/admin/api/routes.go
+++ b/admin/api/routes.go
@@ -62,6 +62,9 @@ func (h *RoutesHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 					Rate1:   tg.Timer.Rate1(),
 					Pct99:   tg.Timer.Percentile(0.99),
 				}
+				if tg.RedirectCode != 0 {
+					ar.Dst = fmt.Sprintf("%d|%s", tg.RedirectCode, tg.URL)
+				}
 				routes = append(routes, ar)
 			}
 		}
--- a/admin/api/routes.go
+++ b/admin/api/routes.go
@@ -62,9 +62,6 @@ func (h *RoutesHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 					Rate1:   tg.Timer.Rate1(),
 					Pct99:   tg.Timer.Percentile(0.99),
 				}
-				if tg.RedirectCode != 0 {
-					ar.Dst = fmt.Sprintf("%d|%s", tg.RedirectCode, tg.URL)
-				}
 				routes = append(routes, ar)
 			}
 		}
--- a/registry/consul/parse_test.go
+++ b/registry/consul/parse_test.go
@@ -47,6 +47,12 @@ func TestParseTag(t *testing.T) {
 			route: "xx/Yy",
 			ok:    true,
 		},
+		{
+			tag:   "p-www.bar.com:80/foo https://www.bar.com/ redirect=302",
+			route: "www.bar.com:80/foo",
+			opts:  "https://www.bar.com/ redirect=302",
+			ok:    true,
+		},
 	}
 
 	for i, tt := range tests {
--- a/registry/consul/parse_test.go
+++ b/registry/consul/parse_test.go
@@ -48,9 +48,9 @@ func TestParseTag(t *testing.T) {
 			ok:    true,
 		},
 		{
-			tag:   "p-www.bar.com:80/foo https://www.bar.com/ redirect=302",
+			tag:   "p-www.bar.com:80/foo redirect=302,https://www.bar.com",
 			route: "www.bar.com:80/foo",
-			opts:  "https://www.bar.com/ redirect=302",
+			opts:  "redirect=302,https://www.bar.com",
 			ok:    true,
 		},
 	}
--- a/registry/consul/service.go
+++ b/registry/consul/service.go
@@ -1,6 +1,7 @@
 package consul
 
 import (
+	"fmt"
 	"log"
 	"net"
 	"runtime"
@@ -119,14 +120,21 @@ func serviceConfig(client *api.Client, name string, passing map[string]bool, tag
 				dst := "http://" + addr + "/"
 				for _, o := range strings.Fields(opts) {
 					switch {
-					case strings.Contains(o, "://"):
-						dst = o
 					case o == "proto=tcp":
 						dst = "tcp://" + addr
 					case o == "proto=https":
 						dst = "https://" + addr
 					case strings.HasPrefix(o, "weight="):
 						weight = o[len("weight="):]
+					case strings.HasPrefix(o, "redirect="):
+						redir := strings.Split(o[len("redirect="):], ",")
+						if len(redir) == 2 {
+							dst = redir[1]
+							ropts = append(ropts, fmt.Sprintf("redirect=%s", redir[0]))
+						} else {
+							log.Printf("[ERROR] Invalid syntax for redirect: %s. should be redirect=<code>,<url>", o)
+							continue
+						}
 					default:
 						ropts = append(ropts, o)
 					}
--- a/registry/consul/service.go
+++ b/registry/consul/service.go
@@ -119,6 +119,8 @@ func serviceConfig(client *api.Client, name string, passing map[string]bool, tag
 				dst := "http://" + addr + "/"
 				for _, o := range strings.Fields(opts) {
 					switch {
+					case strings.Contains(o, "://"):
+						dst = o
 					case o == "proto=tcp":
 						dst = "tcp://" + addr
 					case o == "proto=https":
--- a/route/route.go
+++ b/route/route.go
@@ -6,6 +6,7 @@ import (
 	"net/url"
 	"reflect"
 	"sort"
+	"strconv"
 	"strings"
 
 	"github.com/fabiolb/fabio/metrics"
@@ -68,6 +69,7 @@ func (r *Route) addTarget(service string, targetURL *url.URL, fixedWeight float6
 		t.StripPath = opts["strip"]
 		t.TLSSkipVerify = opts["tlsskipverify"] == "true"
 		t.Host = opts["host"]
+		t.RedirectCode, _ = strconv.Atoi(opts["redirect"])
 	}
 
 	r.Targets = append(r.Targets, t)
@@ -134,7 +136,12 @@ func contains(src, dst []string) bool {
 }
 
 func (r *Route) TargetConfig(t *Target, addWeight bool) string {
-	s := fmt.Sprintf("route add %s %s %s", t.Service, r.Host+r.Path, t.URL)
+	s := fmt.Sprintf("route add %s %s", t.Service, r.Host+r.Path)
+	if t.RedirectCode != 0 {
+		s += fmt.Sprintf(" %d|%s", t.RedirectCode, t.URL)
+	} else {
+		s += fmt.Sprintf(" %s", t.URL)
+	}
 	if addWeight {
 		s += fmt.Sprintf(" weight %2.4f", t.Weight)
 	} else if t.FixedWeight > 0 {
@@ -215,7 +222,7 @@ func (r *Route) weighTargets() {
 	}
 
 	// compute the weight for the targets with dynamic weights
-	dynamic := (1 - sumFixed) / float64(len(r.Targets)-nFixed)
+	dynamic := float64(1-sumFixed) / float64(len(r.Targets)-nFixed)
 	if dynamic < 0 {
 		dynamic = 0
 	}
--- a/route/route.go
+++ b/route/route.go
@@ -76,7 +76,7 @@ func (r *Route) addTarget(service string, targetURL *url.URL, fixedWeight float6
 				log.Printf("[ERROR] redirect status code should be numeric in 3xx range. Got: %s", opts["redirect"])
 			} else if t.RedirectCode < 300 || t.RedirectCode > 399 {
 				t.RedirectCode = 0
-				log.Printf("[ERROR] redirect status code should in 3xx range. Got: %s", opts["redirect"])
+				log.Printf("[ERROR] redirect status code should be in 3xx range. Got: %s", opts["redirect"])
 			}
 		}
 	}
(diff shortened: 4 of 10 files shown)


magiconair:
Thx a lot. A number of users have been asking for this. I’ll have a look tonight.

leprechau:
Yes, we were just discussing doing this last week but it looks like you beat us to it.  Thank you, this is one of the few things we're still using HAProxy for on the frontend.

ctlajoie:
Happy to help! I hope the syntax for indicating a redirect is acceptable. It's not even close to what was discussed in #87 but it enables you to redirect to any given URL which I find valuable.

Also what's up with travis? The failure doesn't seem legit.

leprechau:
Looks like golang `tip` and the Vault checks aren't getting along.

magiconair:
lets disable `tip` for now then. I've got some cycles in the coming weeks to clean out the issue list. So lets get this done and merged.

magiconair:
This is doing the right thing in the `http_proxy.go` but the config implementation is more complex than it needs to be IMO. It should be enough if you add a `redirect=<code>` option since the syntax parsing is mostly there. The notable exception is https://github.com/fabiolb/fabio/blob/master/registry/consul/service.go#L120-L131 which you have covered already and which could use some tests (cough). Then you don't need an `addRedirectTarget` method and can just add that one option to `addTarget` method. 

I assume you need to make the changes for matching the ports in order to do the `http -> https` redirect. I'm wondering whether this wouldn't be more expressive

```
urlprefix-example.com/ https://example.com proto=http redirect=302
urlprefix-example.com/
```

I'd also like to see an integration test in https://github.com/fabiolb/fabio/blob/master/proxy/http_integration_test.go

magiconair:
It may also make sense to get #385 fixed first. I have the patch but need to add the test.

ctlajoie:
Ok I'll see what I can get done on this tonight. Understood regarding tests.. will add those too.


> I assume you need to make the changes for matching the ports in order to do the http -> https redirect. 

Correct. This makes it possible to define a route explicitly on http but not https (and vice versa), which was required for this feature. 

My understanding is that the `proto=` option tells fabio what protocol to use for comms with upstream. I wouldn't recommend using it to define route-matching behavior.

magiconair:
The `proto=` option means whatever we want it to mean. It can mean different things in different contexts if its meaning can be unambiguously derived from the context. Lets get this down to the minimal and tested change first and then we can polish it.

magiconair:
I've just merged #388 so you want to rebase.

ctlajoie:
I implemented your suggestions, minus the `proto=http` one.
In writing the integration test I discovered some issues that I'm not comfortable resolving without feedback. The integration test fails... have a look at it and I think you'll see the problem. It's copying the `Path` part of the request URL over to the target. In many cases this will be desired, but not always. I am considering adding another opt like `absolute=true`. Thoughts?

magiconair:
Got the kids today. So this might be tomorrow.

route/route.go
```diff
@@ -134,7 +136,12 @@ func contains(src, dst []string) bool {
 }
 
 func (r *Route) TargetConfig(t *Target, addWeight bool) string {
-	s := fmt.Sprintf("route add %s %s %s", t.Service, r.Host+r.Path, t.URL)
+	s := fmt.Sprintf("route add %s %s", t.Service, r.Host+r.Path)
+	if t.RedirectCode != 0 {
```
magiconair: This is still using the old syntax

route/route.go
```diff
@@ -215,7 +222,7 @@ func (r *Route) weighTargets() {
 	}
 
 	// compute the weight for the targets with dynamic weights
-	dynamic := (1 - sumFixed) / float64(len(r.Targets)-nFixed)
+	dynamic := float64(1-sumFixed) / float64(len(r.Targets)-nFixed)
```
magiconair: gofmt?

route/table_test.go
```diff
@@ -416,6 +416,9 @@ func TestTableParse(t *testing.T) {
 				targetURLs := make([]string, len(r.wTargets))
 				for i, tg := range r.wTargets {
 					targetURLs[i] = tg.URL.Scheme + "://" + tg.URL.Host + tg.URL.Path
+					if tg.RedirectCode != 0 {
```
magiconair: old syntax here as well

route/table.go
```diff
@@ -273,8 +273,8 @@ func (t Table) route(host, path string) *Route {
 
 // normalizeHost returns the hostname from the request
 // and removes the default port if present.
-func normalizeHost(req *http.Request) string {
-	host := strings.ToLower(req.Host)
+func normalizeHost(host string, req *http.Request) string {
```
magiconair: The reason I've passed in `*http.Request` was that I needed both host and the TLS state. If you pass in the `host` then you might as well pass in the TLS state, e.g.

```go
func normalizeHost(host string, tls bool) string {
    host = strings.ToLower(host)
    if !tls && strings.HasSuffix(host, ":80") {
        return host[:len(host)-len(":80")]
    }
    if tls && strings.HasSuffix(host, ":443") {
        return host[:len(host)-len(":443")]
    }
    return host
}

...

normalizeHost(req.Host, req.TLS != nil)
```

route/target.go
```diff
@@ -33,6 +33,9 @@ type Target struct {
 	// URL is the endpoint the service instance listens on
 	URL *url.URL
 
+	// RedirectCode is the HTTP status code used for redirects.
```
magiconair:
```go
// RedirectCode is the HTTP status code used for redirects.
// When set to a value > 0 the client is redirected to the target url.
```

route/route.go
```diff
@@ -68,6 +69,7 @@ func (r *Route) addTarget(service string, targetURL *url.URL, fixedWeight float6
 		t.StripPath = opts["strip"]
 		t.TLSSkipVerify = opts["tlsskipverify"] == "true"
 		t.Host = opts["host"]
+		t.RedirectCode, _ = strconv.Atoi(opts["redirect"])
```
magiconair: Can we log the error or catch this somewhere sooner and log it there?

registry/consul/parse_test.go
```diff
@@ -47,6 +47,12 @@ func TestParseTag(t *testing.T) {
 			route: "xx/Yy",
 			ok:    true,
 		},
+		{
+			tag:   "p-www.bar.com:80/foo https://www.bar.com/ redirect=302",
+			route: "www.bar.com:80/foo",
+			opts:  "https://www.bar.com/ redirect=302",
```
magiconair: Ah, I forgot that we need to be able to specify the redirect target. How about this?

```
redirect=302,http://www.bar.com/
```

admin/api/routes.go
```diff
@@ -62,6 +62,9 @@ func (h *RoutesHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 					Rate1:   tg.Timer.Rate1(),
 					Pct99:   tg.Timer.Percentile(0.99),
 				}
+				if tg.RedirectCode != 0 {
```
magiconair: old syntax

registry/consul/service.go
```diff
@@ -119,6 +119,8 @@ func serviceConfig(client *api.Client, name string, passing map[string]bool, tag
 				dst := "http://" + addr + "/"
 				for _, o := range strings.Fields(opts) {
 					switch {
+					case strings.Contains(o, "://"):
```
magiconair: This would then also become more obvious by looking for `redirect=xxx`

registry/consul/parse_test.go
ctlajoie: Sure I can do that syntax for the consul service tags.

route/table.go
ctlajoie: Yup, good call.

route/route.go
magiconair: If you divide something by a `float64` then the result can never be an `int` since Go doesn't auto-convert between numeric types. `const` values are the only exception.

ctlajoie: Ok, this makes sense. My bad.

magiconair: No worries. You live and learn :)

route/route.go
```diff
@@ -69,9 +69,15 @@ func (r *Route) addTarget(service string, targetURL *url.URL, fixedWeight float6
...
 		t.TLSSkipVerify = opts["tlsskipverify"] == "true"
 		t.Host = opts["host"]
-		t.RedirectCode, _ = strconv.Atoi(opts["redirect"])
-		if t.RedirectCode < 300 || t.RedirectCode > 399 {
-			t.RedirectCode = 0
+
+		if opts["redirect"] != "" {
+			t.RedirectCode, err = strconv.Atoi(opts["redirect"])
+			if err != nil {
+				log.Printf("[ERROR] redirect status code should be numeric in 3xx range. Got: %s", opts["redirect"])
```
ctlajoie: Ideally these kind of errors should prevent the route from being added, but that would take some refactoring.

magiconair: That could be added the route parser but then it would need to treat opts not as an opaque k/v string. I can have a look at this if and when I write a better parser.

magiconair:
You missed one spot: 

```
route/table_test.go:480:48: cannot use tt.req (type *http.Request) as type bool in argument to normalizeHost
```

ctlajoie:
I made the requested changes but we're not done yet unfortunately. As I mentioned in an earlier comment the integration test fails due to the way the `targetURL` is constructed in http_proxy.go. With 
`route add mock  /foo   http://a.com/abc   opts "redirect=301"`, we get redirected to `http://a.com/foo`. So I was considering implementing some way to indicate you want the target URL to be absolute, in case this behavior is not desired. Can you comment on that?

magiconair:
Yeah, I'm not sure that matters since the target URL should always be absolute. There is no path mangling in fabio since the routing table is usually generated and the `http://host:port/` is really just `host:port`. Originally, I wanted to keep this open and used URLs for that but this may be misleading. So in fabio you can't forward `/foo` to `/bar/foo`. This isn't supported and so far not many have complained about that.

magiconair:
I think the other thing that bugs me a bit but for which I don't have a good answer right now is the continued divergence between the syntax of the `urlprefix-` tags and the `route` commands. 

```
urlprefix-/ redirect=302,http://www.google.com/

# translates to
route add svc / http://www.google.com/ opts "redirect=302"
```

I need to think about this a bit more.

magiconair:
That isn't something that I'd want to address in this PR anyway but I'd like to keep in mind. One approach could be to write a template instead of a special syntax.

```
fabio <templated route add command>

e.g.

fabio route add $svc / $addr opts "redirect=302"
```

Then you'd have to learn only one syntax.

ctlajoie:
I'm not so sure we always want it to be absolute. Suppose you have these routes
```
route add svc example.com:80 / https://example.com redirect=302
route add svc example.com:443 / 1.2.3.4:5678
```
If you hit `http://example.com/some/deep/link` you get redirected to `https://example.com/some/deep/link`. This seems very useful, but I can also see how it might not be desired, depending on the purpose of your redirect route.

magiconair:
How do other load balancers address that?

ctlajoie:
Don't know.. I'll do a bit of research.

magiconair:
Thx. That'd be nice. It's getting a bit late here in NL. Going to hit the pillow since the kids get up early. We're getting there.

leprechau:
I can comment on how it's done with HAProxy where we do our `http` to `https` redirects.

```
    redirect scheme https code 301 if { hdr(Host) -i my.ena.com } !{ ssl_fc }
```

We have that syntax in our main frontend that binds on both 80 and 443 and redirects any request coming in with a `Host` header of  `my.ena.com` to the exact same URI but using the `https` scheme.  That means if I try to access `http://my.ena.com/foo/bar/thing.html` I will get a `301` redirect to `https://my.ena.com/foo/bar/thing.html` with the above configuration.

magiconair:
We could add some magic with a terminating slash or not. `redirect=301,http://foo.com` would append the path from the original request and `redirect=301,http://foo.com/bar` would redirect to `/bar`. That wouldn't solve the path mangling if you want to redirect `/foo` to `/foo/bar` or `/bar/foo`. Another approach could be to add a pseudo-variable like `$path` at the end. This would give you full control over the redirect target.

```
request uri: http://bar.com/foo

redirect=301,http://foo.com$path -> http://foo.com/foo
redirect=301,http://foo.com/ -> http://foo.com/
redirect=301,http://foo.com/bar$path -> http://foo.com/bar/foo
redirect=301,http://foo.com/bar/$path -> http://foo.com/bar/foo # ignore double slashes
```

leprechau:
The pseudo variable would be very flexible and is also less ambiguous than magic based on trading vs non-trailing slash paths.

ctlajoie:
I really like the pseudo-variable idea

ctlajoie:
Pushed changes. Have a close look at http_proxy because I rearranged some code in there since I am no longer using the `targetURL` var as the redirect target.

magiconair:
@ctlajoie I will. This looks decent so far but I'll let it bake for a couple more days since I want to think a bit more about edge cases. If me or you can't think of anything major then I'll merge this early next week (Mon or Tue). Sounds good?

ctlajoie:
Sounds good. I'm actually using it right now in a semi-production environment (accessed by people working in my office) so it's getting some real-world testing.

magiconair:
Even better.

magiconair:
I've merged this to master. I had to update the demo server to handle the comma in the `redirect` option but that was a good time to make this change anyway.

Thanks a lot for this. This has been a long requested feature.

ctlajoie:
@magiconair You might want to know that there was/is a problem with current versions of nomad/consul where if you have `://` in a service tag, nomad is unable to deregister that service. See https://github.com/hashicorp/nomad/issues/3620

Should be fixed in the next nomad release.

magiconair:
@ctlajoie Thanks. That's good to know.

PR #398, opened 2017-12-03
Implement custom noroute html response

### LINKED ISSUE #56: Support custom 404/503 error pages
Hi again :)

It could be awesome to specify file or/and use consul key,
that will be used as an answer of 404 or 503 pages.

`file` or `key value` can be same as HAproxy do:

```
HTTP/1.0 404 NOK^M
Cache-Control: no-cache^M
Connection: close^M
Content-Type: text/html^M
^M
<html>
...
</html>
```

tino:
PR for #56 

- [x] Sign CLA
- [x] Implement functionality for consul backend
- [x] Implement functionality for file backend
- [x] Write tests
- [ ] ~Add default no route HTML? (Like all other servers do, but then styled a bit nicer :)~

My first contribution to a Go project, so let me know what can be done better!

If tests are need, for what should I write them?

Closes #56


tino:
BTW, I added https://github.com/fabiolb/fabio/wiki/Contributing#developing, as it took me quite a while to find out how this works with Go. Might need some rewriting?

magiconair:
Hi Tino, thanks for the contribution. I will have a look right away. Also, thanks for adding the Developing section to the Wiki. Your description is spot on!

tino:
On the mutex part, I copied this from the route/table.go. Why is it needed there and not here?

I held off on multiple custom error pages. In my view if you set an error code in the upstream service, you should also provide the content you want to serve, and fabio should not overwrite it.

This page is just for when the complete service could not be found, and is in sync with the `Proxy.NoRouteStatus` value, hence the – bit clumsy, but clearer then other things I could come up with – name.

magiconair:
Re mutex: I guess you're referring to this:

https://github.com/fabiolb/fabio/blob/master/route/table.go#L47-L56

The reason a mutex is required here is that it changes two things: the table and the metrics registry.

I'll have a look at the rest today.

tino:
Ah okay. I missed that.

tino:
I decided not to add a default HTML message. Although I would like not to serve a browsers default 503/404 by default to humans, it took me too much effort to cleanly implement with templates how I had it in mind.

Is this mergeable as-is, or do you need some more changes? If so, we can start to remove our nginx instances that are in front of fabio atm ☺️

magiconair:
I’ll have a quick look today

magiconair:
This looks fine except that the name is still bugging me because it is too specific. How about we call it `TemplatePath` and keep a generic list of templates. Then for the no route case we use the `noroute.html` template.

I can do this or you can pick this up. I'd like to get this merged as well.

magiconair:
What I suggested is a larger refactor. I'll pick that up myself later. I'm going to merge this and amend the code with some small cleanup. 

Thanks for this!

traceypooh:
i'm...  at an utter loss here...

i finally got consul k/v setup with html, but no matter what i do, it comes out as plaintext, w/
```
  Content-Type: text/plain; charset=utf-8
```

type headers regardless of what I try.

is there some magic "front matter"-like top part of the k/v string i'm supposed to use?
(I haven't been able to get any kind of configuration w/ files, like anything swirling around flailage like this to work on `fabio.properties` either.  so... so far, consul k/v seems my only hope!)

```
registry.file.noroutehtmlpath = /etc/fabio/404.html
registry.static.noroutehtmlpath = "<h1>not found</h1>"
registry.static.noroutehtml = "<h1>not found</h1>"
registry.static.noroutehtmlpath = /etc/fabio/404.html
```

any help vastly appreciated as we're trying to make a bigger move to nomad and fabio...

tristanmorgan:
For me I have an HTML snippet in the consul KV and Fabio returns appropriate headers.

```
$ curl -vv https://consul-esm.service.consul
.....
* Connection state changed (MAX_CONCURRENT_STREAMS == 250)!
< HTTP/2 404 
< content-type: text/html; charset=utf-8
< content-length: 189
< date: Thu, 13 Jan 2022 03:19:01 GMT
< 
<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN" "http://www.w3.org/TR/REC-html40/loose.dtd">
<html><body>
<a href="https://fabio.service.consul/">See Fabio</a>.
* Connection #0 to host consul-esm.service.consul left intact
</body></html>

$ consul kv get fabio/noroute.html
<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN" "http://www.w3.org/TR/REC-html40/loose.dtd">
<html><body>
<a href="https://fabio.service.consul/">See Fabio</a>.
</body></html>

```

traceypooh:
thanks @tristanmorgan ! 

ooh, that totally worked for me.
(i wonder if there's some kind of autodetection for starting kv string with `<!DOCTYPE..` -- but that works while yesterday lots of attempts w/o that lead line was not working for me).

🙏

PR #414, opened 2017-12-16
Identify services using both the ID and the Node

### LINKED ISSUE #383: [bug?] Fabio uses "global" Consul ServiceID's
Fairly serious "bug?" if looking at how Consul implements service registration. By default if you register a Consul service the ServiceID will be set to the ServiceName if not explicitly passed ( https://www.consul.io/api/agent/service.html#id )
This is allowed as a ServiceID only has to be unique per agent, not globally (in the catalog).

Although Consul deals with serviceIDs like that (unique per node) Fabio doesn't, leading to very surprising/disturbing results.. Register a service with the same ServiceName (/ServiceID) on 2 separate nodes. Put 1 of them in maintenance mode (`consul maint -enable -service=<serviceID>`) and **all** routes disappear. Put the entire node into maintenance mode (`consul maint -enable`) and the routes associated with the node do not disappear at all. Let the healthcheck on 1 of them fail and the corresponding service route is not removed..

Is this intentional? If so, we need a big warning somewhere as this can cause very nasty issues.

alvaroaleman:
This aligns Fabio with the Consul docs[0] which state that
a serviceID has to be unique per agent but not cluster wide.

Fixes #383

[0] https://www.consul.io/api/agent/service.html#id

### DIFF
--- a/registry/consul/service.go
+++ b/registry/consul/service.go
@@ -39,7 +39,10 @@ func servicesConfig(client *api.Client, checks []*api.HealthCheck, tagPrefix str
 	// map service name to list of service passing for which the health check is ok
 	m := map[string]map[string]bool{}
 	for _, check := range checks {
-		name, id := check.ServiceName, check.ServiceID
+		// Make the node part of the id, because according to the Consul docs
+		// the ServiceID is unique per agent but not cluster wide
+		// https://www.consul.io/api/agent/service.html#id
+		name, id := check.ServiceName, check.ServiceID+check.Node
 
 		if _, ok := m[name]; !ok {
 			m[name] = map[string]bool{}
@@ -85,7 +88,7 @@ func serviceConfig(client *api.Client, name string, passing map[string]bool, tag
 	for _, svc := range svcs {
 		// check if the instance is in the list of instances
 		// which passed the health check
-		if _, ok := passing[svc.ServiceID]; !ok {
+		if _, ok := passing[svc.ServiceID+svc.Node]; !ok {
 			continue
 		}
 
--- a/registry/consul/passing.go
+++ b/registry/consul/passing.go
@@ -26,15 +26,18 @@ func passingServices(checks []*api.HealthCheck, status []string) []*api.HealthCh
 		// then check whether the agent is still alive and both the
 		// node and the service are not in maintenance mode.
 		for _, c := range checks {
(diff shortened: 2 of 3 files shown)


magiconair:
That looks simple enough. Can you update `registry/consul/passing_test.go`? It is time  for me to also write a test for `service.go`

alvaroaleman:
Updated `passing_test.go`.

Thanks btw for your awesome work and really fast reaction on issues/PRs!

dropje86:
I think in general `c.Node == svc.Node` should be the first thing to evaluate, right? The Node will always be the most specific condition in that loop, while for example `serfHealth` will be hit for every consul agent.

registry/consul/service.go
```diff
@@ -39,7 +39,10 @@ func servicesConfig(client *api.Client, checks []*api.HealthCheck, tagPrefix str
 	// map service name to list of service passing for which the health check is ok
 	m := map[string]map[string]bool{}
 	for _, check := range checks {
-		name, id := check.ServiceName, check.ServiceID
+		// Make the node part of the id, because according to the Consul docs
+		// the ServiceID is unique per agent but not cluster wide
+		// https://www.consul.io/api/agent/service.html#id
+		name, id := check.ServiceName, check.ServiceID+check.Node
```
dropje86: If you want to align with what Consul servers internally do:
`fmt.Sprintf("%s.%s", check.ServiceID, check.Node)`

alvaroaleman:
@dropje86 Does that actually result in Go not evaluating the other conditionals?

@magiconair Is there still something missing from your POV?

dropje86:
@alvaroaleman Although not clearly documented, Go does use "Short circuit evaluation". See here: https://kuree.gitbooks.io/the-go-programming-language-report/content/22/text.html

alvaroaleman:
Interesting, thanks for the link @dropje86!

I've added a second commit that simplifies the conditional logic in `passing.go` and avoids unnecessary evaluations, depending on what @magiconair thinks I can either squash it or put it in a separate PR.

magiconair:
Sorry, got a bit sick. I'll do this probably tomorrow and then try to run a release right away. Time to use goreleaser.

registry/consul/service.go
magiconair: I'll fix that in a subsequent commit.

magiconair:
@alvaroaleman Thanks for the patch. Merged it and made some small changes in subsequent commits.

PR #423, opened 2018-01-23
TCP+SNI support arbitrary large Client Hello

DanSipola:
Use the TLS data to calculate the length of the Client Hello
message so that arbitrary large messages are supported. Previoulsy
messages larger than 1024 bytes failed due to a fixed size buffer.
Note that a Client Hello message must still fit in one TLS record.

The code for parsing the TLS data has been moved around
a bit to ease unit testing.

This is my first attempt using golang so please be aware of that! I don't think
that the performance impact should be significant. I may even improve the
memory footprint per connection since you often don't need as much as 1024 bytes for the
client hello.


proxy/tcp/sni_proxy.go
```diff
@@ -34,27 +36,86 @@ type SNIProxy struct {
...
+	data, err := tlsReader.Peek(9)
+	if err != nil {
+		log.Print("[DEBUG] tcp+sni: TLS handshake failed (failed to peek data)")
+		if p.ConnFail != nil {
+			p.ConnFail.Inc(1)
+		}
+		return err
+	}
+
+	tlsData, err := createClientHelloBuffer(data)
```
magiconair: I think I'd prefer if this function would only determine the size of the buffer and not allocate it. Also, keep it in the `tls_clienthello.go` since this is where the parsing logic for the TLS ClientHello header is. Also, I'd keep `data` instead of `tlsData` since there is no difference.

proxy/tcp/sni_proxy.go
```diff
@@ -34,27 +36,86 @@ type SNIProxy struct {
...
+		log.Printf("[DEBUG] tcp+sni: TLS handshake failed (%s)", err)
 		if p.ConnFail != nil {
 			p.ConnFail.Inc(1)
 		}
 		return err
 	}
-	data = data[:n]
 
-	host, ok := readServerName(data)
+	host, ok := readServerName(tlsData[5:])
```
magiconair: Please add a comment where the `5:` comes from (magic number).

proxy/tcp/sni_proxy_test.go
```diff
@@ -0,0 +1,58 @@
...
+		// bad handshake length
+		//                                  |----- 0 --------|
+		{0x16, 0x03, 0x01, 0x00, 0xaa, 0x01, 0x00, 0x00, 0x00},
+
+		// Fragmentation (handshake larger than record)
+		//                |-  500 ---|      |----- 497 ------|
+		{0x16, 0x03, 0x01, 0x01, 0xF4, 0x01, 0x00, 0x01, 0xf1},
+	}
+
+	for i := 0; i < len(testCases); i++ {
```
magiconair: I really like these test cases. Can you move them to the `tls_clienthello_test.go` since they test the parsing.

When you refactor the method to just return the size of the buffer you can test for that as well. My suggestion is:

```go
tests := []struct {
    name string
    data []byte
    fail bool
    len int
}{
    {
        name: "bad handshake length",
        data: ...,
        fail: true,
    },
    ...
}

for _, tt := range tests {
    t.Run(tt.name, func(t *testing.T) {
        n, err := clientHelloBufferSize(tt.data)
        if tt.fail && err == nil {
            t.Fatal("should fail")
        }
        if got, want := n, tt.len; !tt.fail && got != want {
            t.Fatalf("got buffer size %d want %d", got, want)
        }
    }
}
```

Also I

proxy/tcp/sni_proxy_test.go
```diff
@@ -0,0 +1,58 @@
...
+		_, err := createClientHelloBuffer(testCases[i])
+		if err == nil {
+			t.Logf("Case idx %d did not return an error", i)
+		}
+		assertEqual(err != nil, true)
+	}
+}
+
+func TestCreateClientHelloBufferOk(t *testing.T) {
```
magiconair: You can then probably roll that test case into the previous table driven test.

proxy/tcp/tls_clienthello_test.go
```diff
@@ -0,0 +1,57 @@
+package tcp
+
+import (
+	"encoding/hex"
+	"testing"
+
+	"github.com/fabiolb/fabio/assert"
+)
+
+func TestReadServerNameBadData(t *testing.T) {
```
magiconair: Can you check whether you can also write a table driven test like the one I've outlined above?

DanSipola:
Thanks for the review! I will have a look at it and get it fixed.

DanSipola:
All valid comments. Thanks for that! I hope that I got it right. I'm not used to the github flow so let me know if I should change something.

magiconair:
Thanks!

Issue #168, opened 2016-09-28
support for adding headers


danilobuerger:
Kind of like #110 but for arbitrary headers like HSTS or HPKP, etc.

magiconair:
Do these headers have static or dynamic values, i.e. does the header value depend on the request?

danilobuerger:
I would say this should be static only

jippi:
:+1 on this, for my case static headers would also be sufficient :)

jippi:
Well, static was a lie...

My usecase is `proxy_set_header X-Request-Start "t=${msec}000";` from nginx. which is needed by https://docs.newrelic.com/docs/apm/applications-menu/features/configuring-request-queue-reporting to monitor routing latency :)

killcity:
I would suppose this issue should be called  'Support for custom fields in the HTTP header'? If so, I am also wondering if this will eventually be supported.

Cheers

magiconair:
If the headers need to be dynamic I might be able to re-use the formatting language I'm using for the access logger. In this case I would need to wrap the values in `{}` or something similar to support the `${msec}000` use case. I'll think about that.

magiconair:
I guess they should be configurable per route as well so this needs to become part of the `urlprefix-` tag options.

jippi:
where is this ticket in the OSS pipeline @magiconair ? :)

magiconair:
@jippi there is no pipeline. Generally, I pick up things that people are asking for and for which I have an idea on how to implement it. Adding a static option to add a set of headers to all requests is a no-brainer. But is that really enough? I think I have an idea on how to approach this but for this I need to understand what you need and I'd like to ask you to think a bit outside the box of your specific use-case, if possible.

I think it is reasonable to expect that people would want to modify headers per route. The main challenge with this approach is that if different services announce different options for the same route then the behavior is undefined. As long as that is documented I think I can live with that. This problem will remain for all other features that modify a route. fabio could detect these things and log them to make detection easier.

Also, I'm thinking that there are three operations: `set`, `add` and `del` for setting, adding and deleting headers, e.g.

```
urlprefix-/foo header_set="X-Request-Start=${time_us}" header_add="Strict-Transport-Security: max-age=60; includeSubDomains" header_del=X-Client-IP
```

Does that look like a reasonable approach?

This requires to first update the k/v parser to accept quoted values. Then I'd have to figure out which values would be needed and whether the access logging code could be re-used for that. Then I need to actually modify the headers.

[...]

PR #424, opened 2018-01-23
Add support for enabling HSTS for TLS requests

leprechau:
Support for adding the HSTS header that may help address a small part of #168 and let you get an A+ on Qualys SSL test.


leprechau:
Silly me ... this was added in the wrong spot on the request not the response.

leprechau:
fixed and forced pushed, will create another pr

PR #425, opened 2018-01-24
Add support for HSTS response headers and provide method for adding additional response headers

leprechau:
Support for adding the HSTS header that may help address a small part of #168 and let you get an A+ on Qualys SSL test.

I've deployed this in our DEV and QA environments and verified it is working and adding the HSTS headers as configured.


leprechau:
I put the response header bits in a new function  under `http_headers.go` called `addResponseHeders` and added a new `TestAddResponseHeaders` to `http_headers_test.go`.  Currently the STS headers are the only things being added via the `addResponseHeaders` function but I assume that could change in the future.

magiconair:
Interesting. I assume you tested this since I thought you'd need to overwrite the `ModifyResponse` function in the reverse proxy. An integration test might be handy. If this works then I can refactor the ugly hack that I have for getting the response body size and status code. 

Could you please update `config/load_test.go` with the new arguments and add the docs as well? Then I'll merge this.

Thx

leprechau:
Yes, I have this running in a dev environment and the STS header is added to the response.

I’ll get the docs and extra tests going today.

-- Aaron

leprechau:
Okay, the docs, configuration tests and integration test are now committed.

PR #426, opened 2018-01-24
Add option to allow Fabio to register frontend services in Consul on behalf of user services

rileyje:
We would like to have Fabio re-register itself in Consul under user-specified names, so that users can access their services through Fabio via Consul-provided A records. The motivation for this is:

1. Users can immediately access their service "foo" directly at foo.service.example.com using Consul DNS
2. Users do not need to create CNAMES to Fabio for their service in an external DNS system, e.g. foo.example.com IN CNAME fabio.service.example.com
3. The service will be reachable via a DNS health-checked A record so that Kerberos auth in browsers will work correctly with a Kerberos principal unique to the user's service name. Currently this fails if the service is accessed via a CNAME to Fabio, unless principals are made too generically to match (e.g.) fabio.service.example.com.

The intended usage is that users will give their service a backend-specific name, and pass to Fabio the name they wish to use for the front end via a new "alias" option. For example:

        "ServiceName": "foo-backend",
        "ServiceTags": [
            "urlprefix-foo.service.example.com/ alias=foo",
            "urlprefix-foo/"
        ],

Fabio then registers new service foo.service.example.com with its own host and port information and creates routes "foo.service.example.com/" and "foo/" to foo-backend.service.example.com as usual.


registry/backend.go
```diff
@@ -2,10 +2,13 @@ package registry
 
 type Backend interface {
 	// Register registers fabio as a service in the registry.
-	Register() error
+	Register(serviceName string) error
```
magiconair: I think a better approach would be to just extend the register function since you need to keep the state of what is registered in the backend anyway. 

```go
type Backend interface {
    Register(name ...string) error
    DeregisterAll() error
    ...
}
```

registry/consul/backend.go
```diff
@@ -36,26 +37,47 @@ func NewBackend(cfg *config.Consul) (registry.Backend, error) {
 	return &be{c: c, dc: dc, cfg: cfg}, nil
 }
 
-func (b *be) Register() error {
-	if !b.cfg.Register {
+func (b *be) Register(serviceName string) error {
```
magiconair: Then here you first unregister names that no longer exist and register new names.

registry/consul/backend.go
```diff
@@ -36,26 +37,47 @@ func NewBackend(cfg *config.Consul) (registry.Backend, error) {
...
+	b.dereg[serviceName] = register(b.c, service)
+
 	return nil
 }
 
-func (b *be) Deregister() error {
-	if b.dereg != nil {
-		b.dereg <- true // trigger deregistration
-		<-b.dereg       // wait for completion
+func (b *be) Deregister(serviceName string) error {
```
magiconair: This function then deregisters all currently registered services.

registry/consul/backend.go
```diff
@@ -100,6 +122,54 @@ func (b *be) WatchNoRouteHTML() chan string {
 	return html
 }
 
+func (b *be) UpdateAliases(config string) error {
```
magiconair: Move the parsing code into the `route/parse_new.go` since there you already have all the parse helpers. Return a list of alias names to register, e.g.

```go
func ParseAliases(config string) (names []string, error) {...}
```

main.go
```diff
@@ -396,6 +397,8 @@ func watchBackend(cfg *config.Config, first chan bool) {
 			continue
 		}
 
+		registry.Default.UpdateAliases(next)
```
magiconair: Then here you call

```go
aliases, err := route.ParseAliases(config)
registry.Default.Register(aliases)
```

rileyje:
Thanks @magiconair. I tried to address your feedback with these changes:

- removed UpdateAliases() and moved parse code to ```ParseAliases()``` in ```route/parse_new.go```. Implements a subset of ```Parse()```, so there is some redundancy. Better to just call ```Parse()```?. Had to import ```config``` to check if fabio should register itself. Can't get that from the private config member of the backend. Brought in ```log``` for some visibility.
- Added ```DeregisterAll()```, but kept ```Deregister()``` for single deregistrations. I wasn't sure if that's what you had in mind.
- Extended ```Register()``` to handle removal and addition of aliases.
- Changed option name to ```register``` (much better :)
- Still tracking registered services in ```b.dereg``` map. Is that OK? Wasn't clear on "you need to keep the state of what is registered in the backend anyway"

I had a couple of questions while working on this:

- Would you be OK with adding "DeregisterCriticalServiceAfter" to registrations so that if fabio exits uncleanly, Consul will clean up services? That was added in Consul 0.7.0, not sure if you have a minimum support Consul version or we could check version with the API before including that option.
- Any concerns about duplicate health checks? Would be nice if Consul allowed you to specify an existing health check when registering a service. I can open a feature request against Consul.
- Do you have any thoughts on test coverage for ```Register()``` and ```Deregister(All)()```?

I see the Travis CI build failed, I'll look in to that.

leprechau:
The Consul API does allow you to pass existing checks via the `AgentServiceRegistration` struct via the `Checks` element which references `AgentServiceChecks` which is a slice of pointers to `AgentServiceCheck`.

leprechau:
You may be able to register existing services and checks with `CatalogRegistration` but I'm not sure that has the intended result.  I'll check locally and see how it works.

registry/consul/backend.go
```diff
@@ -36,25 +36,67 @@ func NewBackend(cfg *config.Consul) (registry.Backend, error) {
...
-	b.dereg = register(b.c, service)
 	return nil
 }
 
-func (b *be) Deregister() error {
-	if b.dereg != nil {
-		b.dereg <- true // trigger deregistration
-		<-b.dereg       // wait for completion
+func (b *be) Deregister(serviceName string) error {
+	if len(serviceName) == 0 {
```
magiconair: Since `DeregisterAll()` exists I wouldn't use a magic value `""` to deregister all services. Every now and then I am tempted to do that but then opt for simplicity. `Deregister(service)` deregisters a service if it is currently registered, otherwise it does nothing. `DeregisterAll()` deregisters all currently registered services.

registry/consul/backend.go
```diff
@@ -36,25 +36,67 @@ func NewBackend(cfg *config.Consul) (registry.Backend, error) {
...
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	// register new service names
+	for _, serviceName := range services {
+		if _, ok := b.dereg[serviceName]; ok {
+			log.Printf("[DEBUG]: %q already registered (appears in dereg map)\n", serviceName)
```
magiconair: pls drop the text in `(...)` since that is an implementation detail.

registry/consul/backend.go
```diff
@@ -36,25 +36,67 @@ func NewBackend(cfg *config.Consul) (registry.Backend, error) {
...
+		}
+	}
+
+	return nil
+}
+
+func (b *be) DeregisterAll() error {
+	log.Printf("[DEBUG]: consul: Deregistering all registered Fabio aliases.")
+	for name, dereg := range b.dereg {
+		if dereg != nil {
```
magiconair: I prefer less indents

```go
for name, dereg := range b.dereg {
    if dereg == nil {
        continue
    }
    ...
}
```

registry/consul/backend.go
```diff
@@ -36,25 +36,67 @@ func NewBackend(cfg *config.Consul) (registry.Backend, error) {
...
+		if !stringInSlice(serviceName, services) {
+			err := b.Deregister(serviceName)
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	// register new service names
+	for _, serviceName := range services {
```
magiconair: less indents

```go
for _, name := range services {
    if b.dereg[name] == nil {
        continue
    }
    ...
}
```

registry/consul/register.go
```diff
@@ -38,10 +38,11 @@ func register(c *api.Client, service *api.AgentServiceRegistration) chan bool {
 
 	register := func() string {
 		if err := c.Agent().ServiceRegister(service); err != nil {
-			log.Printf("[ERROR] consul: Cannot register fabio in consul. %s", err)
+			log.Printf("[ERROR] consul: Cannot register fabio [name:%q] in Consul. %s", service.Name, err)
 			return ""
 		}
 
+		log.Printf("[INFO] consul: Registered fabio with name %q", service.Name)
```
magiconair: `Registered fabio as %q`

route/parse_new.go
```diff
@@ -87,6 +90,42 @@ func Parse(in string) (defs []*RouteDef, err error) {
 	return defs, nil
 }
 
+func ParseAliases(in string, cfg *config.Config) (names []string, err error) {
```
magiconair: Pls add a comment on what this function does.

route/parse_new.go
```diff
@@ -87,6 +90,42 @@ func Parse(in string) (defs []*RouteDef, err error) {
...
+		}
+		if err != nil {
+			return nil, fmt.Errorf("line %d: %s", i+1, err)
+		}
+		defs = append(defs, def)
+	}
+
+	var aliases []string
+
+	if cfg.Registry.Backend == "consul" && cfg.Registry.Consul.Register {
```
magiconair: Why is that important here? I'd just parse the aliases and let the backend decide what to do with them.

route/parse_new.go
```diff
@@ -87,6 +90,42 @@ func Parse(in string) (defs []*RouteDef, err error) {
...
+		aliases = append(aliases, cfg.Registry.Consul.ServiceName)
+	}
+
+	for _, d := range defs {
+		registerName, ok := d.Opts["register"]
+		if ok {
+			log.Printf("[DEBUG]: consul: Service %q wants alias %q", d.Service, registerName)
```
magiconair: I don't think we should log that here.

magiconair:
I think adding `DeregisterCriticalServiceAfter` with a sane default makes sense. Can you please add that as a separate commit within this PR?

magiconair:
Ah, and please add some tests to https://github.com/fabiolb/fabio/blob/master/route/parse_test.go

rileyje:
@magiconair here is the latest version of the patch.
- Updated ```Commands``` const to included ```register```. I noticed that ```host``` was missing and took the liberty of adding it.
- Updated docs to include ```register``` option.
- Removed config check from ```ParseAliases()``` and moved it to Consul backend
- Added comment for ```ParseAliases()```
- Added tests for ```ParseAliases()```
- Removed unneeded magic value from ```Deregister()```
- Factored out excess indents
- Cleaned up logging statements
- Added ```DeregisterCriticalServiceAfter``` as a separate commit. Used the 90m default from Consul docs. Not sure if that fits your definition of sane :) We will likely set it to something much shorter in production.

magiconair:
I think this is good enough as is. Thanks a lot!

rileyje:
Great, thank you!

Issue #439, opened 2018-02-09
Feature Request: IP Whitelisting


microadam:
Apologies if this is mentioned elsewhere, I couldn't see anything.

It would be great if we could whitelist specific hostnames to specific IP ranges. Not sure if something like that is planned or outside of the scope of this project?

Thanks!

magiconair:
No this is t supported right now but shouldn’t be too hard to add. How often do these ip ranges change?

Also, no need to apologize :)

microadam:
Good to know is relatively easy! I wouldn't have thought they would change TOO often. Being easily updatable is always nice though. Possibly another `urlprefix-` option?

leprechau:
I was working on a PR for this exact thing or very similar.  A few options to allow simple ACL type behavior.

Example:

only allow traffic from the specified source blocks (default deny)
`urlprefix-/foobar/ allow=172.16.0.0/12,10.0.0.0/8,192.168.0.0/16`

allow all traffic except traffic from the specified source blocks (default allow)
`urlprefix-/foobar/ deny=1.0.1.0/24,1.0.2.0/23`

leprechau:
@microadam would that address your need?  @magiconair have any suggestions around that approach?

magiconair:
That looks good. The only question I had what to do with the PROXY connections? You'd have to use the `RemoteAddr` field from the `http.Request` to make the decision but that gets overwritten by `go-proxyproto`. I'm not sure whether this is an issue. Just something to think about.

leprechau:
That's true, it wouldn't be the actual source at that point but the "declared" source from `proxyproto`.  I think that's addressable via documentation unless you feel otherwise.  The other option would be to preserve the original remote address separate from `proxyproto` but I haven't looked how much structural change that would entail.

magiconair:
@leprechau IPv6 addresses, hostnames (dns reverse lookup timeout comes to mind...)

I also thought about more complex use cases like testing for cookies and so forth but I think that should live in the app layer. 

While thinking about this: Is this something that should be configured on a per-route level or on a global level? What is the underlying use-case?

leprechau:
In our use case we have some services that need to be exposed via the load balancer but shouldn't necessarily have complete unrestricted access.  We could use firewall rules on the load balancer itself but that would require dedicated IPs or bind ports for those specific services.  Having a light ACL option on the route would allow this to happen without needing dynamic listeners.

leprechau:
[...]

PR #442, opened 2018-02-13
add basic ip centric access control on routes

leprechau:
I've verified and tested this locally with good results.  Running it in our development environment for testing as well.

### DIFF
--- a/docs/content/cfg/_index.md
+++ b/docs/content/cfg/_index.md
@@ -24,14 +24,16 @@ Add a route for a service `svc` for the `src` (e.g. `/path` or `:port`) to a `ds
 
 `route add <svc> <src> <dst>[ weight <w>][ tags "<t1>,<t2>,..."][ opts "k1=v1 k2=v2 ..."]`
 
-Option               | Description
--------------------- | -----------
-`strip=/path`        | Forward `/path/to/file` as `/to/file`
-`proto=tcp`          | Upstream service is TCP, `dst` must be `:port`
-`proto=https`        | Upstream service is HTTPS
-`tlsskipverify=true` | Disable TLS cert validation for HTTPS upstream
-`host=name`          | Set the `Host` header to `name`. If `name == 'dst'` then the `Host` header will be set to the registered upstream host name
-`register=name`      | Register fabio as new service `name`. Useful for registering hostnames for host specific routes.
+Option                                     | Description
+------------------------------------------ | -----------
+`allow=ip:10.0.0.0/8,ip:192.168.0.0/24`    | Restrict access to source addresses within the `10.0.0.0/8` or `192.168.0.0/24` CIDR mask.  All other requests will be denied.
+`deny=ip:10.0.0.0/8,ip:1.2.3.4/32`         | Deny requests that source from the `10.0.0.0/8` CIDR mask or `1.2.3.4`.  All other requests will be allowed.
+`strip=/path`                              | Forward `/path/to/file` as `/to/file`
+`proto=tcp`                                | Upstream service is TCP, `dst` must be `:port`
+`proto=https`                              | Upstream service is HTTPS
+`tlsskipverify=true`                       | Disable TLS cert validation for HTTPS upstream
+`host=name`                                | Set the `Host` header to `name`. If `name == 'dst'` then the `Host` header will be set to the registered upstream host name
+`register=name`                            | Register fabio as new service `name`. Useful for registering hostnames for host specific routes.
 
 ##### Example
 
--- a/docs/content/cfg/_index.md
+++ b/docs/content/cfg/_index.md
@@ -26,8 +26,8 @@ Add a route for a service `svc` for the `src` (e.g. `/path` or `:port`) to a `ds
 
 Option                                     | Description
 ------------------------------------------ | -----------
-`allow=ip:10.0.0.0/8,ip:192.168.0.0/24`    | Restrict access to source addresses within the `10.0.0.0/8` or `192.168.0.0/24` CIDR mask.  All other requests will be denied.
-`deny=ip:10.0.0.0/8,ip:1.2.3.4/32`         | Deny requests that source from the `10.0.0.0/8` CIDR mask or `1.2.3.4`.  All other requests will be allowed.
+`allow=ip:10.0.0.0/8,ip:fe80::/10`         | Restrict access to source addresses within the `10.0.0.0/8` or `fe80::/10` CIDR mask.  All other requests will be denied.
+`deny=ip:10.0.0.0/8,ip:fe80::1234`         | Deny requests that source from the `10.0.0.0/8` CIDR mask or `fe80::1234`.  All other requests will be allowed.
 `strip=/path`                              | Forward `/path/to/file` as `/to/file`
 `proto=tcp`                                | Upstream service is TCP, `dst` must be `:port`
 `proto=https`                              | Upstream service is HTTPS
--- a/docs/content/feature/access-control.md
+++ b/docs/content/feature/access-control.md
@@ -1,0 +1,31 @@
+---
+title: "Access Control"
+since: "1.5.8"
+---
+
+fabio supports basic ip centric access control per route.  You may
+specify one of `allow` or `deny` options per route to control access.
+Currently only source ip control is available.
+
+<!--more-->
+
+To allow access to a route from clients within the `192.168.1.0/24`
+and `10.0.0.0/8` subnet you would add the following option:
+
+```
+allow=ip:192.168.1.0/24,ip:10.0.0.0/8
+```
+
+With this specified only clients sourced from those two subnets will
+be allowed.  All other requests to that route will be denied.
+
+
+Inversely, to only deny a specific set of clients you can use the
+following option syntax:
+
+```
+deny=ip:1.2.3.4/32,100.123.0.0/16
+```
+
+With this configuration access will be denied to any clients with
+the `1.2.3.4` address or coming from the `100.123.0.0/16` network.
--- a/docs/content/feature/access-control.md
+++ b/docs/content/feature/access-control.md
@@ -35,3 +35,15 @@ Single host addresses (addresses without a prefix) will have a
 That means `1.2.3.4` is equivalent to `1.2.3.4/32` and `fe80::1234`
 is equivalent to `fe80::1234/128` when specifying
 address blocks for `allow` or `deny` rules.
+
+The source ip used for validation against the defined ruleset is
+taken from information available in the request.
+
+For `HTTP` requests the client `RemoteAddr` is always validated
+followed by the first element of the `X-Forwarded-For` header, if
+present.  When either of these elements match an `allow` the request
+will be allowed; similarly when either element matches a `deny` the
+request will be denied.
+
+For `TCP` requests the originating address of the network socket
+is used as the sole paramater for validation.
--- a/docs/content/feature/access-control.md
+++ b/docs/content/feature/access-control.md
@@ -10,22 +10,28 @@ Currently only source ip control is available.
 <!--more-->
 
 To allow access to a route from clients within the `192.168.1.0/24`
-and `10.0.0.0/8` subnet you would add the following option:
+and `fe80::/10` subnet you would add the following option:
 
 ```
-allow=ip:192.168.1.0/24,ip:10.0.0.0/8
+allow=ip:192.168.1.0/24,ip:fe80::/10
 ```
 
 With this specified only clients sourced from those two subnets will
 be allowed.  All other requests to that route will be denied.
 
 
-Inversely, to only deny a specific set of clients you can use the
+Inversely, to deny a specific set of clients you can use the
 following option syntax:
 
 ```
-deny=ip:1.2.3.4/32,100.123.0.0/16
+deny=ip:fe80::1234,100.123.0.0/16
 ```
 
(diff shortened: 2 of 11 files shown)


leprechau:
Oops, forgot to link this to #439

route/access_rules.go
```diff
@@ -0,0 +1,128 @@
...
+func (t *Target) AccessDeniedHTTP(r *http.Request) bool {
+	host, _, err := net.SplitHostPort(r.RemoteAddr)
+
+	if err != nil {
+		log.Printf("[ERROR] Failed to get host from RemoteAddr %s: %s",
+			r.RemoteAddr, err.Error())
+		return false
+	}
+
+	// prefer xff header if set
```
leprechau: I have some mixed feelings about this section.  In the case where fabio may be behind another proxy you would want to honor `XFF` but then what about the actual originator?  Would it be better to check both the `XFF` (if present and different) as well as the actual source and return `true` from this function if either the `XFF` or `RemoteAddr` pass returns `true`?

route/access_rules.go
```diff
@@ -0,0 +1,127 @@
...
+	if xff := r.Header.Get("X-Forwarded-For"); xff != "" && xff != host {
+		host = xff
+	}
+
+	// currently only one function - more may be added in the future
+	return t.denyByIP(net.ParseIP(host))
+}
```
leprechau: I have mixed feelings about this section.  In the event that fabio is deployed behind another load balancer you would want to check the `XFF` header but then what about the actual source?  Would it be better to check both the `XFF` (if present and different) as well as the actual `RemoteAddr` and return `true` (deny) if either of those checks return `true` on their own?

leprechau:
It seems the tests are randomly failing with a timeout around some consul interaction.  I've seen the different go versions succeed and fail at different times with the same code by just amending the commit and doing force push.

route/access_rules.go
leprechau: Yes, I definitely don't like this.  It's an open exploit as is since there is no validation on who is allowed to pass an `XFF` header to fabio.

Example with `allow=ip:172.16.0.0/12` and a client that is not sourcing from that network.

```
$ curl -H 'X-Forwarded-For: 172.16.1.1' https://test-site/foo
{}
$ curl https://test-site/foo
access denied
```

I think it needs to check both `XFF` and `RemoteAddr` or it's basically useless.

leprechau:
I've fixed the security hole in the latest commit.

leprechau:
@magiconair I think that's it unless you have other suggestions or requests.

magiconair:
I haven't fully nailed down the test failures. I'm punting by running on travis and codeship and if one of them is green I'm ok with that :-/ I'll have a look at the PR later.

leprechau:
@microadam are you able to test and verify?

microadam:
I will see if I can give it a go! Not a go developer, but will give compiling it a shot

leprechau:
@microadam I could post a binary at my fork of the repo if that makes it easier.  Not sure how comfortable you would be with that though ... and I completely understand.

microadam:
@leprechau yeah that would definitely work! Hitting lots of errors trying to compile (probably just me not knowing how to do it!)

leprechau:
Linux amd64?

microadam:
Have managed to compile it! will try and test it out

leprechau:
Okay, awesome ... let me know if you need anything else.

microadam:
I assume I must be doing something wrong, but with options of `allow=172.14.0.0/12` and accessing it from an address of 192.168.0.x, I don't get denied, which is what I was expecting

I am on Mac OSX running fabio on a Vagrant box, so there are a few networks to contend with

leprechau:
Interesting ... can you turn on access logging just to see what IP fabio sees your connection coming from?  There is probably a NAT between you and the vbox host.  To be really sure you could switch it to  `deny=0.0.0.0/0` that would block everything.

microadam:
No joy with the deny all either. I can only assume I have compiled it wrong. Perhaps if you could provide a binary after all?

Steps to compile were:

```
git clone git@github.com:myENA/fabio.git
git checkout feature/route-acl
go get ./...
make linux
```
I assume something in there is not quite right!

microadam:
Think I have it compiled correctly now. Getting this error:

`[ERROR] failed to process access rules: invalid access item, expected <type>:<data>, got [0.0.0.0/0]`

My `urlprefix` looks like: `urlprefix-my-site.com/ deny=0.0.0.0/0`

magiconair:
@microadam and @leprechau thanks to both of you for working on this!

leprechau:
Also, just realized my own example up there was wrong ... sorry about that.

microadam:
Finally got around to doing some tests, apologies for the delay!

A few things I have spotted, some are possibly edge cases and not worth worrying about (possibly all).

* with `deny=ip:0.0.0.0/0` I can still access it from localhost. Request comes through like:
```
[::1] - - [14/Feb/2018:20:38:10 +0000] "GET / HTTP/1.1" 200 68744
```

* If it put in `allow=ip:192.168.0.252` I correctly get:
```
[ERROR] failed to process access rules: failed to parse CIDR ip:192.168.0.252 with error: invalid CIDR address: 192.168.0.252
```
However it then defaults to allowing all traffic through. As we don't have any concept of "validation" before the server runs, perhaps this should prevent the server from running, or at least, if its an `allow` rule, it should block EVERYTHING, at least you then know something is wrong? Not sure how to handle the case of where its a `deny` rule though, as you wouldn't want to default to ALLOWING everything as you wouldn't know there was an issue. Just trying to think of the case where this is all automated and people aren't paying attention to the logs.

Seems to work in all the other trivial tests I have done! I haven't managed to test it behind another proxy yet, as I don't have that easily setup, but from the looks of the code it should handle that!

Thanks a lot for doing this :)

leprechau:
Thanks for the feedback.  Internally the code is using the golang core `net` package.  Specifically it's using the ParseCIDR and Contains functions.

https://golang.org/pkg/net/#ParseCIDR
https://golang.org/pkg/net/#IPNet.Contains

They support both IPv4 and IPv6 but will not transit boundaries.  So, for the first example with `deny=ip:0.0.0.0/0` that will match ALL IPv4 addresses but will not match IPv6 addresses.  The `[::1]` request source is IPv6 localhost.

The second part is also a side affect of the ParseCIDR function.  `192.168.0.252` is not in proper CIDR notation.  To specify a single IP address in CIDR form it would be `192.168.0.252/32` (a 32 bit mask denotes a single address).

To stopping server operation and/or ignoring the route I have mixed feelings.  The current behavior for invalid options goes ahead and adds the route but just ignores the bad options.  I kept his same behavior for the `allow` and `deny` options.  It is completely possible to either skip the route (not add it) or stop the server if route parsing fails.  However, since routes may be added dynamically I don't think stopping the server for an invalid route option makes any sense.

microadam:
Sure. That all makes sense. 

Just trying to think if there is a way to make it really obvious that it has failed without checking the logs. There isn’t really a clean way of doing it I suppose. I guess if you do mess it up and don’t notice the logs, you will just not see the change you were expecting I.e person still won’t have access or the traffic you are trying to block will still come through. I guess that is the best that can be done!

Definitely a nice addition to this project, certainly covers my use case :)

leprechau:
Awesome, thank you for the feedback and testing.

@magiconair any other thoughts or concerns?  What do you think about skipping the route on failure for allow/deny processing?  Just putting a `continue` in the loop if the process function returns error?

magiconair:
I’ll have a look at the code later after coffee. One thought I had about the addresses is whether to automatically change 1.2.3.4 to 1.2.3.4/32. 

Also, could we add an `all` alias which will just match any address?

Can you do `allow=ip:1.2.3.4/32 deny=all` or is that implicit?

leprechau:
You could check for a string contains and add the slash 32 when there isn’t a match.  I’m not sure an all is needed with the current behavior.  The code currently only allows one of ‘allow’ or ‘deny’ to be specified.  When using the allow option all sources not matched by the allow are denied (default deny).  When using the ‘deny’ option all sources not matched by the deny are allowed (default allow).

leprechau:
I added some code to add a `/32` prefix to value strings that do not contain a `/` ... invalid or incomplete addresses will still error out in the ParseCIDR function.

magiconair:
Should we do this for v6 addresses without netmask as well?

—
Frank Schröder

leprechau:
That's a good question ... would require a bit more pre-parsing but I think it's doable and probably good to be consistent.

leprechau:
Latest commit should properly add a `/32` prefix to plain IPv4 addresses and a `/128` prefix to plain IPv6 addresses.

leprechau:
Last commit added equal test coverage for IPv4 and IPv6 addresses.

docs/content/cfg/_index.md
```diff
@@ -24,14 +24,16 @@ Add a route for a service `svc` for the `src` (e.g. `/path` or `:port`) to a `ds
...
--------------------- | -----------
-`strip=/path`        | Forward `/path/to/file` as `/to/file`
-`proto=tcp`          | Upstream service is TCP, `dst` must be `:port`
-`proto=https`        | Upstream service is HTTPS
-`tlsskipverify=true` | Disable TLS cert validation for HTTPS upstream
-`host=name`          | Set the `Host` header to `name`. If `name == 'dst'` then the `Host` header will be set to the registered upstream host name
-`register=name`      | Register fabio as new service `name`. Useful for registering hostnames for host specific routes.
+Option                                     | Description
+------------------------------------------ | -----------
+`allow=ip:10.0.0.0/8,ip:192.168.0.0/24`    | Restrict access to source addresses within the `10.0.0.0/8` or `192.168.0.0/24` CIDR mask.  All other requests will be denied.
```
magiconair: Pls add an IPv6 example as well

docs/content/feature/access-control.md
```diff
@@ -0,0 +1,31 @@
+---
+title: "Access Control"
+since: "1.5.8"
+---
+
+fabio supports basic ip centric access control per route.  You may
+specify one of `allow` or `deny` options per route to control access.
+Currently only source ip control is available.
```
magiconair: Pls add an ipv6 example and explain how the source ip is determined. Also add a comment that `/32` and `/128` is added automatically.

proxy/http_proxy.go
```diff
@@ -84,6 +84,11 @@ func (p *HTTPProxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if t.AccessDeniedHTTP(r) {
+		http.Error(w, "access denied", http.StatusForbidden)
```
magiconair: I'm pretty sure that someone will ask for a different status code here but we can leave it for now.

proxy/tcp/sni_proxy.go
```diff
@@ -78,6 +78,11 @@ func (p *SNIProxy) ServeTCP(in net.Conn) error {
 	}
 	addr := t.URL.Host
 
+	if t.AccessDeniedTCP(in) {
+		log.Print("[INFO] route rules denied access to ", t.URL.String())
```
magiconair: This should also log who was denied access.

proxy/tcp/tcp_proxy.go
```diff
@@ -48,6 +48,11 @@ func (p *Proxy) ServeTCP(in net.Conn) error {
 	}
 	addr := t.URL.Host
 
+	if t.AccessDeniedTCP(in) {
+		log.Print("[INFO] route rules denied access to ", t.URL.String())
```
magiconair: same here

leprechau:
I had to export the `ProcessAccessRules` function that hangs off of `route.Target` in `access_rules.go` in order to process the rules for the integration test.

I also moved the deny logging into the `denyByIP` function in `access_rules.go` to make it easier to alter if needed in the future.

docs/content/feature/access-control.md
leprechau: done!

proxy/http_proxy.go
leprechau: I can see that being a request but not sure there is much more appropraite than forbidden for a denied request.  Would you want that to be a config option at some point?

magiconair: yeah, if someone asks for this.

docs/content/feature/access-control.md
```diff
@@ -0,0 +1,49 @@
...
+Single host addresses (addresses without a prefix) will have a
+`/32` prefix, for IPv4, or a `/128` prefix, for IPv6, added automatically.
+That means `1.2.3.4` is equivalent to `1.2.3.4/32` and `fe80::1234`
+is equivalent to `fe80::1234/128` when specifying
+address blocks for `allow` or `deny` rules.
+
+The source ip used for validation against the defined ruleset is
+taken from information available in the request.
+
+For `HTTP` requests the client `RemoteAddr` is always validated
```
magiconair: I think if the inbound connection uses the PROXY protocol then this is mapped into the `RemoteAddr` field of the underlying connection. This should be the case for both TCP and HTTP since this is on the connection itself.

https://github.com/fabiolb/fabio/blob/master/vendor/github.com/armon/go-proxyproto/protocol.go#L96-L113

I'll update the docs before merging.

magiconair:
hrmpf, I've merged this but put the additional commit into the branch. This way github won't auto-close the PR ... This is on master now. 

Thanks a lot for @leprechau and @microadam for working on this!

PR #449, opened 2018-02-26
Check if XFF is available before denying on IP

atillamas:
When using allow=ip:x.x.x.x behind an application loadbalancer Fabio will always deny access since it does the denyByIP before checking if there is a "X-Forwarded-For" available and therefore uses the loadbalancers ip as source, this PR puts the denyByIP after the XFF check and makes it work as intended.

### DIFF
--- a/route/access_rules.go
+++ b/route/access_rules.go
@@ -29,11 +29,6 @@ func (t *Target) AccessDeniedHTTP(r *http.Request) bool {
 		log.Printf("[WARN] failed to parse remote address %s", host)
 	}
 
-	// check remote source and return if denied
-	if t.denyByIP(ip) {
-		return true
-	}
-
 	// check xff source if present
 	if xff := r.Header.Get("X-Forwarded-For"); xff != "" {
 		// only use left-most element (client)
@@ -48,6 +43,11 @@ func (t *Target) AccessDeniedHTTP(r *http.Request) bool {
 				return true
 			}
 		}
+	}
+
+	// check remote source and return if denied
+	if t.denyByIP(ip) {
+		return true
 	}
 
 	// default allow


leprechau:
@atillamas That is intentional.  Trusting the XFF header without also validating the source IP makes adding rules pointless.

Think of this scenario:

You have an `allow=ip:127.0.0.1/8` because you want to restrict a route to localhost.  If you do not validate both source and XFF header anyone on the internet can do: `curl -H 'X-Forwarded-For: 127.0.0.1` and gain access to your route.

The only way to make it secure is to trust both the XFF header you expect AND the actual upstream load balancers.  Meaning you will need at least two allows.

route/access_rules.go
```diff
@@ -50,6 +45,11 @@ func (t *Target) AccessDeniedHTTP(r *http.Request) bool {
 		}
 	}
```
leprechau: Moving this block doesn't do what you think.  You have orphaned the actual source IP parsing above making this function do the same thing as the XFF check block.

route/access_rules.go
```diff
@@ -29,11 +29,6 @@ func (t *Target) AccessDeniedHTTP(r *http.Request) bool {
 		log.Printf("[WARN] failed to parse remote address %s", host)
 	}
 
-	// check remote source and return if denied
```
leprechau: This block belongs with the block above.  The `ip` that is being checked here is the result of the parse of the remote address in the previous block.

atillamas:
@leprechau Hi. Thank you for the info. But using 1.5.8 i can still access a rule that has allow=ip:10.0.0.0/16,ip:X.X.X.X and then using curl -H 'X-Forwarded-For: X.X.X.X' hostname  or even any of the 10.0.0.0/16 adresses give me access where ip from the curl host is something completely different. Should i just close this PR and open an Issue ticket?

leprechau:
The allow rule should enforce that the remote address and the XFF, if present, be inside the allowed mask(s).

leprechau:
The test cases should cover that.  There are several scenarios around XFF and the remote source address.  Keep in mind that the code currently only inspects the first member of the XFF header as per spect that should be the nearest upstream peer.

https://github.com/fabiolb/fabio/blob/master/route/access_rules_test.go#L167-L202

atillamas:
@leprechau Ok, i think I've found the problem.. Apparently AWS decides to append the ip of the client to the end of the XFF list (this is for classic loadbalancer but it seems to apply to Application Loadbalancer also: https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/x-forwarded-headers.html), so if i do:
curl -H 'X-Forwarded-For: 87.9.15.28' host

printing xff variable give me this:

87.9.15.28, 80.13.7.2

is the X-Forwardeded for header set by curl
and 80.13.7.2 is the ip of the client doing the curl

The culprit is here:
https://github.com/fabiolb/fabio/blob/master/route/access_rules.go#L39-L40

leprechau:
Right, that bit of code should be correct.  The leftmost element of the field should be the client.

https://en.wikipedia.org/wiki/X-Forwarded-For#Format

The behavior of proxies is to add/set the XFF header to the remote address (client) if it is not present or append the header if it already exists.

That should mean that if you allow the address(s) of the ALB and the address(s) of the client(s) you want to be able to access the route it should work.

Maybe I am missing something here.

leprechau:
More specifically if you have this sequence of events:

Client (1.2.3.4) -> Proxy (3.4.5.6) -> fabio (7.8.9.0) -> The XFF at fabio should be 1.2.3.4,3.4.5.6 and you would need to allow both the client 1.2.3.4 and the immediate prior proxy 3.4.5.6

The reason ... fabio needs to know who it can trust in this scenario and not blindly accept the XFF header.  You can use the XFF header to allow clients but we also need to validate the remote peer (the proxy) so we don't accept XFF headers from unknown parties.

Make sense?

atillamas:
@leprechau No, since it's the rightmost value that is the true client IP, I can manipulate the leftmost value with the curl -H 'X-Forwarded-For: whateveripiwant'
and whateveripiwant will be the leftmost value and then fabio will let it pass, even though it should be blocked...
This is in AWS behind a LB

leprechau:
That is because you are tricking the ALB by sending an XFF header with curl.  If you do not pass an XFF header with curl the XFF at the ALB should only be the client.  The ALB (like most proxies) will not replace an existing XFF it will simply append it.

atillamas:
Well if "tricking" it with just adding a Header with an allowed IP is the only thing needed to bypass it i wouldn't consider it secure at all.
If this indeed is the intended behaviour I'll continue to stick with WAF.

leprechau:
So you are saying you can bypass the allow list?

atillamas:
Yes, that's what i've been saying all along

leprechau:
Sorry, I wasn't following that.

atillamas:
Wasn't very clear maybe but here is the post again:

@leprechau Hi. Thank you for the info. But using 1.5.8 i can still access a rule that has allow=ip:10.0.0.0/16,ip:X.X.X.X and then using curl -H 'X-Forwarded-For: X.X.X.X' hostname or even any of the 10.0.0.0/16 adresses give me access where ip from the curl host is something completely different. Should i just close this PR and open an Issue ticket?

atillamas:
> Right, that bit of code should be correct. The leftmost element of the field should be the client.
https://en.wikipedia.org/wiki/X-Forwarded-For#Format

Well i guess AWS doesn't follow the standard then:
https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/x-forwarded-headers.html

> If a request from a client already contains an X-Forwarded-For header, Elastic Load Balancing appends the IP address of the client at the end of the header value. In this case, the last IP address in the list is the IP address of the client. For example, the following header contains two IP addresses added by the client, which might not be trustworthy, plus the client IP address added by Elastic Load Balancing:

> X-Forwarded-For: ip-address-1, ip-address-2, client-ip-address

leprechau:
XFF header semantics aside ... what is the IP of the ALB that sits before fabio?  What does fabio see as the "source" of that connection?

leprechau:
Just looking at the code a client should have to fall through both checks in order to be allowed.  The only exception would be an error in the first part of the `AccessDeniedHTTP` function.

If this error is printed the request would be allowed ...
https://github.com/fabiolb/fabio/blob/master/route/access_rules.go#L22-L26

Moving on the next check is for the actual remote address ... this only returns on a deny ...
https://github.com/fabiolb/fabio/blob/master/route/access_rules.go#L22-L26

Next we have the whole XFF parsing and checking ... this also only returns on a deny ...
https://github.com/fabiolb/fabio/blob/master/route/access_rules.go#L22-L26

Lastly we return false if and only if the previous two sections did not return ....
https://github.com/fabiolb/fabio/blob/master/route/access_rules.go#L53-L54

I'll setup some local test cases and see what if anything I can pinpoint.

leprechau:
With stock 1.5.8 with a very basic setup I cannot reproduce this.

Here is my setup .. all running locally on one machine:

Terminal 1:
```
go get github.com/hashicorp/http-echo
http-echo -text="hello"
```

Terminal 2:
```
consul agent -dev
```

Terminal 3:
```
fabio -log.level=TRACE
```

Terminal 4:
```
$ curl --request PUT --data 'route add http-echo / http://127.0.0.1:5678 opts "allow=ip:10.0.0.0/8,ip:127.0.0.1"' 127.0.0.1:8500/v1/kv/fabio/config
true
$ curl http://127.0.0.1:9999/
hello
$ curl --interface 127.0.0.1 -H "X-Forwarded-For: 1.2.3.4" http://127.0.0.1:9999/
access denied
$ curl --interface 127.0.0.2 http://127.0.0.1:9999/
access denied
$ curl --interface 127.0.0.2 --header "X-Forwarded-For: 10.0.0.1" http://127.0.0.1:9999/
access denied
```

This is the expected behavior.  What am I missing here?

leprechau:
Actually ... I think I see the problem with multiple proxies in this case ...
```
$ curl --interface 127.0.0.1 --header "X-Forwarded-For: 10.0.0.2" http://127.0.0.1:9999/
hello
```

Hrmm... so you need to validate the source in this case the upstream proxy but you also need to validate the header.  However when you have an upstream proxy and you are forced to allow it you let anyone on the net to spoof the header.

leprechau:
What we really need is an 'allow xff from' type setting.  Does that exist within the ALB?

atillamas:
> XFF header semantics aside ... what is the IP of the ALB that sits before fabio? What does fabio see as the "source" of that connection?

The Loadbalancer connects from random ips from the subnets specified for the clients. for example 
10.0.60.132
10.0.61.12

Here's some tcpdumps on port 9999 on the fabio client with the request headers:

`curl echo.staging.budbee.com`
```
GET / HTTP/1.1
X-Forwarded-For: 67.209.115.14
X-Forwarded-Proto: http
X-Forwarded-Port: 80
Host: echo.staging.budbee.com
X-Amzn-Trace-Id: Root=1-5a946e06-bde747b0116cf72811e9e6a2
User-Agent: curl/7.29.0
Accept: */*

HTTP/1.1 200 OK
Content-Length: 25
Content-Type: text/plain; charset=utf-8
Date: Mon, 26 Feb 2018 20:28:54 GMT
X-App-Name: http-echo
X-App-Version: 0.2.3
```
`curl -H 'X-Forwarded-For: 67.209.115.14' echo.staging.budbee.com`
```
GET / HTTP/1.1
X-Forwarded-For: 67.209.115.14, 178.132.77.232
X-Forwarded-Proto: http
X-Forwarded-Port: 80
Host: echo.staging.budbee.com
X-Amzn-Trace-Id: Root=1-5a946e22-f9626e80398045b8c1ae4376
User-Agent: curl/7.55.1
Accept: */*

HTTP/1.1 200 OK
Content-Length: 25
Content-Type: text/plain; charset=utf-8
Date: Mon, 26 Feb 2018 20:29:22 GMT
X-App-Name: http-echo
X-App-Version: 0.2.3
```
`curl echo.staging.budbee.com`
```
GET / HTTP/1.1
X-Forwarded-For: 178.132.77.232
X-Forwarded-Proto: http
X-Forwarded-Port: 80
Host: echo.staging.budbee.com
X-Amzn-Trace-Id: Root=1-5a946e29-a4a8610d9e7ac71d926d84f2
User-Agent: curl/7.55.1
Accept: */*

HTTP/1.1 403 Forbidden
Content-Type: text/plain; charset=utf-8
X-Content-Type-Options: nosniff
Date: Mon, 26 Feb 2018 20:29:29 GMT
Content-Length: 14
```
It would probably be easiest to just check if there is a X-Amzn-Trace-Id Header present, and if so, use the rightmost XFF header, (even though I've understood that you don't want to do this). But it would make this whole thing work.

leprechau:
> It would probably be easiest to just check if there is a X-Amzn-Trace-Id Header present, and if so, use the rightmost XFF header, (even though I've understood that you don't want to do this). But it would make this whole thing work.

It's not about a want ... it's about actually enforcing the given ruleset.

We could add some odd if header Z then look at this part of the XFF else look at somewhere else ... but that's just adding more layers of obfuscation and not actually preventing the actual problem when you have another proxy in-front of fabio.

The problem being: in the current implementation when you have another proxy before fabio you are open to arbitrary bypass of the access list by simply sending the right XFF header.

The reason I made the `AccessDeniedHTTP` function check both the actual source as well as the XFF header was to try and prevent this arbitrary bypass.  However, in the case of another proxy before fabio you are forced to blindly trust the source of the proxy which then means you are blindly trusting arbitrary XFF headers.

Trusting client side XFF headers in general is a big security hole in my opinion.  Sane load balancers should replace ... not append ... the XFF unless they are operating in a completely trusted environment.   If you have the ability to strip client sent XFF on the ALB ... force the ALB to set, not append ... that would be your best option now.

leprechau:
Currently fabio does not add the XFF header itself but rather allows the upstream core function do that work.  It respects client sent XFF which is the real problem with using it for access control in my opinion.

https://github.com/golang/go/blob/master/src/net/http/httputil/reverseproxy.go#L186-L194

I think another bool option to fabio like `proxy.trust.xff` that would strip the XFF from the request before it enters the proxy.  This could default to `false` for backwards compatibility but allow it to be set to `true` in cases where there is not another proxy before fabio and fabio wants to ensure unverified client sent XFF are not interpreted or passed on to other applications.

leprechau:
@magiconair do you have any thoughts on the above proposed option?  I could submit a quick PR to handle that if you agree.

leprechau:
@atillamas the other option may be to validate ALL components of the XFF instead of just an arbitrary first or last element.  In the case of an `allow` rule all parts of the XFF plus the actual remote address would have to match for the request to be passed onto the backend.  Similarly for a `deny` rule if any part of the XFF or the actual remote address matched the request would be denied.

I think this would be acceptable.  Thoughts?

leprechau:
What do you think of this as a replacement for the current logic in the XFF part of the `AccessDeniedHTTP` function:

```go
// check xff source if present
if xff := r.Header.Get("X-Forwarded-For"); xff != "" {
	// Trusting xff headers sent from clients is dangerous and generally
	// bad practice.  Therefore, we cannot assume which if any of the elements
	// is the actual client address.  To try and avoid the chance of spoofed
	// headers and/or loose upstream proxies we validate all elements in the header.
	for xip := range strings.Split(xff, ",") {
		// ensure we only get the ip string
		xip = strings.TrimSpace(xip)
		// only continue if xip differs from host
		if xip != host {
			if ip = net.ParseIP(xip); ip == nil {
				log.Printf("[WARN] failed to parse xff address %s", xip)
			}
			// check xff source and return if denied
			if t.denyByIP(ip) {
				return true
			}
		}
	}
}
```

leprechau:
See the proposed "fix" in the #453 referenced above.

atillamas:
>  It's not about a want ... it's about actually enforcing the given ruleset.
> 
> We could add some odd if header Z then look at this part of the XFF else look at somewhere else ... but that's just adding more layers of obfuscation and not actually preventing the actual problem when you have another proxy in-front of fabio.
> 
> The problem being: in the current implementation when you have another proxy before fabio you are open to arbitrary bypass of the access list by simply sending the right XFF header.

I just don't get it. We can trust the headers given by AWS Loadbalancers, atleast the **LAST ENTRY** it will be appended by AWS and will always be the client IP connecting to the loadbalancer, it is to be trusted as the source of the connection and cannot be manipulated. As long as you trust AWS you can trust that value.

> @atillamas the other option may be to validate ALL components of the XFF instead of just an arbitrary first or last element. In the case of an allow rule all parts of the XFF plus the actual remote address would have to match for the request to be passed onto the backend. Similarly for a deny rule if any part of the XFF or the actual remote address matched the request would be denied.
> 
> I think this would be acceptable. Thoughts?

Yes this would work for our usecase  atleast. But as soon as someone needs to actually use XFF headers it will become a problem i would assume?

leprechau:
Validating all elements shouldn’t break anyone needing to use XFF for a layered proxy setup.  It also ensures that we are not exclusively trusting client sent headers.

I know that in this particular instance it’s the last element.  But the last element just means the client to the proxy before fabio.  I don’t want to assume that there will never be a case where there is more than one hop before fabio.  This wouldn’t be a good design but I can’t really control what others deploy.

In all cases I can imagine validating all the elements of the header seems to be the best option without having any control over what the upstream peer, or more importantly the internet facing peer, is sending.

atillamas:
>I know that in this particular instance it’s the last element. But the last element just means the client to the proxy before fabio. I don’t want to assume that there will never be a case where there is more than one hop before fabio. This wouldn’t be a good design but I can’t really control what others deploy.

Isn't that how it's supposed to work? That's how firewall rules work. You take the source ip and base the rules of that. Is it allowed, is it denied.
Anyways. Don't feel like it's any point in discussing this further.
Thanks for taking the time with this.
Should i just close this PR?

leprechau:
Yes, the last element should always be the closest peer to where you are reading the header.  However, like I said just reading the last element doesn't ensure it's the client address ... it could be the address of the proxy before the proxy before fabio.

Proxies behind proxies is just kind of a mess and should really be avoided if at all possible.  I think the solution in #453 will address the case of any number of proxies before fabio and handle them in the best way possible.

To this PR, yes you can close it if you're happy with the solution in the #453 PR.

atillamas:
@leprechau 
Yes I'm happy with it.
Thanks for your effort.
Closing this

magiconair:
As a side note people have been saying to put an nginx in front of fabio to handle cases like generic redirects, https enforcements and header mangling since fabio does not support this. If this would again sit behind an ELB you have the multi-layered proxy setup.

leprechau:
Correct, I have another PR coming to address the blind trust of client XFF headers in cases where fabio is indeed the internet facing authoritative load balancer (my personal deployment).

leprechau:
Also not directly related but I think fabio is quickly addressing those use cases.  It now supports generic redirects, https enforcement and some header injection.  There are also quite a few discussions around the structure for full on header manipulation that could be implemented.

We've been tracking fabio for quite some time and over the last few months have decided all of our needs are addressed.  We have fabio deployed to production for a small set of applications and have plans to fully replace our haproxy/consul-template setup in the near future.

Thank you!

PR #451, opened 2018-02-26
Catch all redirect

magiconair:
This patch adds the option to redirect to the same host but maybe on a different path or scheme.

For this `$host` can be used on the target URL, e.g.

```
route add https-redirect / https://$host$path opts "redirect=301"
```

### DIFF
--- a/route/target.go
+++ b/route/target.go
@@ -78,5 +78,8 @@ func (t *Target) GetRedirectURL(requestURL *url.URL) *url.URL {
 			redirectURL.RawQuery = requestURL.RawQuery
 		}
 	}
+	if strings.Contains(redirectURL.Host, "$host") {
+		redirectURL.Host = strings.Replace(redirectURL.Host, "$host", requestURL.Host, 1)
+	}
 	return redirectURL
 }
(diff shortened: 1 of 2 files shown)


magiconair:
Hmm, not sure whether this creates a redirect loop.

leprechau:
If we interpret that `https-redirect` is only valid for originating `http` that should be redirected to `https` I think this works.  Meaning that `https-redirect` will never match an incoming `https` request.

Otherwise we could keep the same `http-redirect` with the additional `$host` pseudo var and restrict on port in the rule something like...
```
route add http-redirect :80 https://$host$path opts "redirect=301"
```

leprechau:
Sorry, just looked at the code ... there's not a new option it is just adding the pseudo var.  So I think that you would get a redirect loop with a path `/` match but shouldn't with a port `:80` match.

magiconair:
Something like `route add :80/ ...` might be handy. Not sure if that fits well with the current syntax

leprechau:
Agreed a host agnostic match with a port and path is exactly what you would want for this specific use case.

holtwilkins:
How about https://github.com/fabiolb/fabio/pull/544 instead?

PR #456, opened 2018-02-27
add ability to strip client xff header - see #449

leprechau:


proxy/http_proxy.go
```diff
@@ -84,6 +84,15 @@ func (p *HTTPProxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// If we are the internet facing load balancer
+	// we should not accept a client XFF header.
+	// We should only accept an XFF header here if we
+	// are running in a trusted environment behind another
+	// properly configured load balancer.
+	if !p.Config.TrustXFF {
+		r.Header.Del("X-Forwarded-For")
```
magiconair: Would you need to remove the other `X-Forwarded-*` headers as well?

docs/content/ref/proxy.header.trust.xff.md
```diff
@@ -0,0 +1,16 @@
+---
+title: "proxy.header.trust.xff"
+---
+
+`proxy.header.trust.xff` instructs fabio to trust the X-Forwarded-For header
```
magiconair: Maybe backtick `X-Fowarded-For` and `true` and `false`? I need to work on the CSS a bit more though.

magiconair:
If we add the ability to remove arbitrary untrusted headers then this would become an improvement instead of a potential breaking change.

magiconair:
Along the same line we could also add the option of injecting headers.

magiconair:
But if the XFF problem is your main sticking point then we can start with removal.

leprechau:
Generic header trust is definitely something I am concerned with and would love to address in a more elegant fashion.  The main reason for this pull request is to address what I see as broken behavior in the core golang reverse proxy (and AWS).

In my opinion the internet facing authoritative load balancer should not blindly append and pass an XFF header from an untrusted client.  There are probably other headers that we should allow/deny on a per header and possibly even per route instance.  The most obvious of those to me however is the XFF header.

leprechau:
I can see fabio eventually having something similar to the `header_upstream` and `header_downstream` directives available in Caddy (https://caddyserver.com/docs/proxy).  These allow the user to control per header weather to add/set, append, or remove headers coming from the client before being passed to the backend and similarly filtering headers coming from a backend before being sent to the client.

leprechau:
Closing this in favor of a more robust header manipulation system.

Issue #452, opened 2018-02-26
glob matching improvements


sharbov:
Hey,

We've been using fabio for a while now, and we recently started using the glob matcher.

The problem we've encountered was that once we switched the matching strategy it affected all the endpoints matching, in addition to that the glob matcher does not match the '/' signs. these two issues forced us to add many new tags to existing services in order to try and cover all of the possibilities of each endpoint 

e.g

```
urlprefix-/service/foo/*
urlprefix-/service/foo/*/*
urlprefix-/service/foo/*/*/* ...
```

This is clearly not the preferred solution, I would like to suggest three ways to go about this issue:

* Add a per tag matching strategy (e.g `urlglob-/service/foo/*`).
* Add support for double '*' wild card, which will match '/' signs.
* Add a new matcher -  a regex matcher.

If any of these ideas sounds reasonable to you, I would be happy to open a pull request for them.

relates to https://github.com/fabiolb/fabio/issues/341

Thanks,
Shay

magiconair:
Hey Shay,

I think extending the current glob matcher with `**` logic would be a good idea and it should be backwards compatible. 

Adding a regex matcher in addition to that would help, too. 

Bonus points for adding a route option to select this per route instead of using a global flag. 

PR is very welcome. 
Frank
—
Frank Schröder

eliasbrange:
Just stumbled upon this and installed `1.5.9` to fix the problem. However, now it seems to work with both `*` and `**`. E.g. `/some/path/here` is matched by both `/**` and `/*`. What are the preferred way?

PR #457, opened 2018-02-28
Add new & improved glob matcher

sharbov:
I've added a new matcher using `gobwas/glob` side-by-side to the existing glob matcher

This might hinder performance due to the fact that the matcher is compiling the pattern each time.

I was thinking of using a cache for the compiled matchers (something like `patrickmn/go-cache`).

@magiconair what do you think?

https://github.com/fabiolb/fabio/issues/452


sharbov:
Hey @magiconair,

I've implemented the fixes you've requested.
As far as I can tell the two glob packages are compatible, 
I used the same test to validate it (and extended it to test the new capabilities).

route/matcher_test.go
```diff
@@ -55,35 +60,6 @@ func TestGlobMatcher(t *testing.T) {
...
-	for _, tt := range tests {
-		t.Run(tt.uri, func(t *testing.T) {
-			if got, want := gobwasGlobMatcher(tt.uri, tt.route), tt.matches; got != want {
-				t.Fatalf("got %v want %v", got, want)
-			}
-		})
-	}
-}
\ No newline at end of file
+func getRoute(path string) *Route {
```
magiconair: I'd prefer if you compile the glob matcher in the test since this keeps the changeset smaller, e.g.

```go
for _, tt := range tests {
    t.Run(tt.uri, func(t *testing.T) {
        tt.route.Matcher = glob.MustCompile(tt.route.path)
        ...
    })
}
```

route/route.go
```diff
@@ -36,6 +37,9 @@ type Route struct {
 	// total contains the total number of requests for this route.
 	// Used by the RRPicker
 	total uint64
+
+	// Matcher represents compiled pattern.
+	Matcher glob.Glob
```
magiconair: I think `Glob` is a better name than `Matcher`.

route/table.go
```diff
@@ -14,6 +14,7 @@ import (
 
 	"github.com/fabiolb/fabio/metrics"
 	"github.com/ryanuber/go-glob"
+	glob2 "github.com/gobwas/glob"
```
magiconair: Please name this `glob` and also remove the `github.com/ryanuber/go-glob` library in a separate commit.

route/table.go
```diff
@@ -153,7 +154,7 @@ func (t Table) addRoute(d *RouteDef) error {
 	switch {
 	// add new host
 	case t[host] == nil:
-		r := &Route{Host: host, Path: path}
+		r := &Route{Host: host, Path: path, Matcher: glob2.MustCompile(path)}
```
magiconair: What happens if this is an invalid pattern? Does it `panic`? You probably need to catch the error and return it.

route/table.go
```diff
@@ -290,7 +298,8 @@ func (t Table) matchingHosts(req *http.Request) (hosts []string) {
 	host := normalizeHost(req.Host, req.TLS != nil)
 	for pattern := range t {
 		normpat := normalizeHost(pattern, req.TLS != nil)
-		if glob.Glob(normpat, host) {
```
sharbov: I'm not sure why you used the glob matcher here in the first place, can a host contain wildcards ?

magiconair: Good question. I'll check.

magiconair: Yes, you can have a route for `*.example.com/foo`

magiconair: I wrote a small benchmark to verify performance. `gobwas` is fast but only when the expression is pre-compiled. It is only 1.4 ms but this sits in the hot-path. However, there is no easy way to refactor this since `Table` is just a `map[string]Routes`. Might be a good time to finally change this to a struct. I'll have a look.

```
$ go test -bench .
goos: darwin
goarch: amd64
pkg: globbench
BenchmarkGlob/ryanuber-8         	20000000	        91.4 ns/op
BenchmarkGlob/gobwas-8           	 1000000	      1385 ns/op
BenchmarkGlob/gobwas-precompile-8         	200000000	         7.29 ns/op
PASS
ok  	globbench	5.537s
```

```go
package main

import "testing"
import ryanuber "github.com/ryanuber/go-glob"
import gobwas "github.com/gobwas/glob"

var n int

func BenchmarkGlob(b *testing.B) {
	b.Run("ryanuber", func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			if ryanuber.Glob("*.example.com", "foo.example.com") {
				n++
			}
		}
	})
	b.Run("gobwas", func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			m := gobwas.MustCompile("*.example.com")
			if m.Match("foo.example.com") {
				n++
			}
		}
	})
	b.Run("gobwas-precompile", func(b *testing.B) {
		m := gobwas.MustCompile("*.example.com")
		for i := 0; i < b.N; i++ {
			if m.Match("foo.example.com") {
				n++
			}
		}
	})
	b.Log(n)
}
```

PR #466, opened 2018-03-13
Check upstream X-Forwarded-Proto prior to redirect

### LINKED ISSUE #448: Redirect http to https on the same destination
Hi, is it possible to use fabio to redirect http to https on the same destination? for example redirect http://example.com to https://example.com ? when I've tried it i just get a redirect-loop.

leprechau:
The AWS load balancer terminates SSL and passes plain HTTP
to fabio but does not offer the ability to do a schema redirect.

Therefore, fabio running behind an AWS ELB with a schema
redirect rule results in a redirect loop. 

Fixes #448

### DIFF
--- a/proxy/http_proxy.go
+++ b/proxy/http_proxy.go
@@ -99,8 +99,7 @@ func (p *HTTPProxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if t.RedirectCode != 0 {
-		redirectURL := t.GetRedirectURL(requestURL)
-		http.Redirect(w, r, redirectURL.String(), t.RedirectCode)
+		http.Redirect(w, r, t.RedirectURL.String(), t.RedirectCode)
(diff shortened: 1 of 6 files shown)


magiconair:
Thanks a lot!

proxy/http_proxy.go
```diff
@@ -99,8 +99,7 @@ func (p *HTTPProxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if t.RedirectCode != 0 {
-		redirectURL := t.GetRedirectURL(requestURL)
```
leprechau: @magiconair I know this already got merged but I was thinking maybe the above `if` statement should be `t.RedirectCode != 0 && t.RedirectURL != nil` just to be safe?

magiconair: Agreed. See #477

PR #468, opened 2018-03-18
Fix the regex of the example proxy.gzip.contenttype

tino:
The backslash wasn't properly escaped, making fabio not wanting to start when uncommenting the example in `fabio.properties`


magiconair:
Thanks!

Issue #347, opened 2017-09-08
%20 in route is causing route mismatch,  regression in 1.5.2 , works with 1.3.7


devendram:
our route with % in it used to work fine with 1.3.7 . After upgrading to 1.5.2 we are hitting issue with such routes . Please see following from fabio log

log
2017/09/08 12:44:17 [WARN] No route for localhost/run/before/test%201

route exists 
host	    localhost/run/before/test%201	http://127.0.0.1:22275/		100.00%

Tcpdump with fabio 1.3.7
=================

GET /run/before/test%201 HTTP/1.1
Host: localhost

HTTP/1.1 200 OK
Content-Length: 222
Content-Type: text/plain; charset=utf-8
Date: Fri, 08 Sep 2017 19:37:55 GMT
......

Tcpdump with fabio 1.5.2
===================

GET /run/before/test%201 HTTP/1.1
Host: localhost

HTTP/1.1 404 Not Found
Date: Fri, 08 Sep 2017 19:29:52 GMT
Content-Length: 0
Content-Type: text/plain; charset=utf-8

magiconair:
Is the `%20` part of the route or the request?

```
rotue add svc /x/a%20b http://1.2.3.4:5678

or

rotue add svc /x http://1.2.3.4:5678

```

magiconair:
Another difference is fabio 1.3.7 was built with go1.7.5 and fabio 1.5.2 is built with go1.8.3. Could be something in the go http package but I'm looking.

magiconair:
nvm, I've found it. This is a consequence of #219 which uses `req.URL.Path` instead of `req.RequestURI` for the lookup. The first one is parsed whereas the other one isn't. Since you can encode a space in two ways in a URL (`%20` and `+`) I think the new way is the correct one but then the entries in the routing table need to be URL decoded. I need to think about this a bit. Is there something you need from 1.5.2 and this is a blocker?

jshaw86:
@magiconair we upgraded to 1.5.2 for better TLS+SNI support which we need for one of our backend microservices(SOC Compliance).

jshaw86:
@magiconair any idea on effort on this? We are trying to decide whether or not to invest our resources into this issue or not.

magiconair:
How urgent is this since a) I'm sick right now and b) need to get something ready for HashiConf next week. Can this wait a couple of days?

jshaw86:
[...]

Issue #486, opened 2018-04-19
Fabio decodes URL path parameters


mcneiljt:
Related issue/change:
https://github.com/fabiolb/fabio/issues/347
https://github.com/fabiolb/fabio/commit/cc3e3cd601abbb48a6f0d152a45c9c013b78b399

While using fabio a team mate noticed that some of our URL path parameters, such as resource Ids, which have url-encoded content in them are being decoded by fabio causing 404 errors. The decoded path parameters are sent to the service and not recognized as a valid path.

Example:
container/job with `urlprefix-/posts`
request: GET /posts/abcder%2Fsdf394/comments

This matches the URL prefix, but routes to `/posts/abcder/sdf394/comments` which isn't a valid path and thus returns a 404.

tuempeltaucher:
Any chance to see a fix here? We have the same problem with https://domain.tld as a path param.

leprechau:
There is work and discussion going on in #489 ... would you be able to test if that patch fixes your issue?

tuempeltaucher:
No, that didn't change anything.

[TRACE] abc Tracing xxx/stella/v1/xzy/https://wwwi.screenwork.de/
[...]
[TRACE] abc Routing to service stella on http://10.0.1.4:25853/
"GET /stella/v1/xyz/https%3A%2F%2Fwww.screenwork.de%2F HTTP/1.1" 404 43 "" "curl/7.29.0" http://10.0.1.4:25853/v1/xyz/https://www.screenwork.de/

But I traced it done and added a few "RawPath"s:

https://github.com/tuempeltaucher/fabio/commit/a06f98503d063e88b123176296f79a03bf0bf3ea

leprechau:
@tuempeltaucher that's with the code from #489 applied?

tuempeltaucher:
Yes, I forked valentin-krasontovitsch:url_decoding_in_routing, that is the branch from his pull request.

tuempeltaucher:
Any news here?

PR #489, opened 2018-04-25
Feat: Pass encoded characters in path unchanged

valentin-krasontovitsch:
We enable passing substrings that represent encoded reserved symbols, like
`%2f` for `/`, in the path of a request.

NB: This does not enable using encoded characters in rewrite rules (e.g. in the
urlprefix directive).

Cf. #347 and #486

### DIFF
--- a/route/target.go
+++ b/route/target.go
@@ -70,20 +70,37 @@ func (t *Target) BuildRedirectURL(requestURL *url.URL) {
 		Scheme:   t.URL.Scheme,
 		Host:     t.URL.Host,
 		Path:     t.URL.Path,
+		RawPath:  t.URL.Path,
 		RawQuery: t.URL.RawQuery,
 	}
+	// treat case of $path not separated with a / from host
 	if strings.HasSuffix(t.RedirectURL.Host, "$path") {
 		t.RedirectURL.Host = t.RedirectURL.Host[:len(t.RedirectURL.Host)-len("$path")]
 		t.RedirectURL.Path = "$path"
 	}
+	// remove / before $path in redirect url
 	if strings.Contains(t.RedirectURL.Path, "/$path") {
 		t.RedirectURL.Path = strings.Replace(t.RedirectURL.Path, "/$path", "$path", 1)
+		t.RedirectURL.RawPath = strings.Replace(t.RedirectURL.RawPath, "/$path", "$path", 1)
 	}
+	// insert passed request path, remove strip path, set quer
 	if strings.Contains(t.RedirectURL.Path, "$path") {
+		// replace in not raw path
 		t.RedirectURL.Path = strings.Replace(t.RedirectURL.Path, "$path", requestURL.Path, 1)
+		// replace in raw path - determine replacement first
+		var replaceRawPath string
+		if requestURL.RawPath == "" {
+			replaceRawPath = requestURL.Path
+		} else {
+			replaceRawPath = requestURL.RawPath
+		}
+		t.RedirectURL.RawPath = strings.Replace(t.RedirectURL.RawPath, "$path", replaceRawPath, 1)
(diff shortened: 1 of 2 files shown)


valentin-krasontovitsch:
So it turns out that one can utilize the internal mechanisms of `net/url` to handle encoded parts of an uri. One just has to keep track of `RawPath`, whenever necessary, which is what this commit attempts to do...

As a side note, I got carried away and tried enabling encoded characters in rewrite rules as well, and in the strip prefix, which worked fine, until you combined say a rule like

```
route add svc / http://bar.com/a%2fc/$path
```

with a request for 
```
/%20
``` 
the space in the request won't get escaped, as it's unambiguous on its own (I guess, cf. [docs for function that sets `Path` and `RawPath` in `net/url` package](https://golang.org/src/net/url/url.go?s=26390:26435#L633)), but would get escaped in the context of `%2f`, i.e. if one looked at the whole url `http://bar.com/a%2fc/%20`.

One may inspect the more involved changes [on my corresponding branch](https://github.com/valentin-krasontovitsch/fabio/tree/fix_all_the_decoding) - in particular, a link to [the failing test case](https://github.com/valentin-krasontovitsch/fabio/blob/003382edb26c4821b53913f26893054c4fb9a9fb/route/target_test.go#L107).

Since I didn't see an easy and quick fix for that, I decided that it's maybe not the most needed use-case and that supporting what we have here might be enough.

magiconair:
Does that mean this fixes #347 and #486?

valentin-krasontovitsch:
I'm so sorry, saw your comment, but kinda ignored it O_o - let me see, I honestly don't know anymore, gotta check out the code... did I add some test cases? : )

... yes, I did. Please take a look at those, I'm sure you're more familiar with the code base and can say with more confidence which cases I have covered, and whether I forgot any cases / should add some cases.

But at least the test cases are covered! ... and I think, to answer your original question, yes, that solves those issues, as far as I understand.

scalp42:
Any chance to see this merged @magiconair ?

makepanic:
We were also hit with this issue. 
Is there anything one can do to proceed with the issue/PR?

leprechau:
@valentin-krasontovitsch Sorry for the lag in response around this issue.  If you could resolve the conflicts I'll get this merged.  Otherwise I'll try and fix the simple conflict myself when I get time.

valentin-krasontovitsch:
No worries, I’ll try to take a look this weekend, if you don’t hear from
me, feel free to attempt a rebase or just patch it directly yourself - I
can’t really say that I recall, but I assume it wasn’t a complicated
change...

valentin-krasontovitsch:
okay, ran a rebase, and luckily it was only a superficial conflict - a test added in the same place where i added a test. resolved, and pushed - after the travis build checks out, you will hopefully be able to merge.

valentin-krasontovitsch:
Works on go1.13, but a handful tests break on go master? Very strange.
I presume somebody more familiar with the code base might have a better chance of working out why the tests fail... I can try to take a look myself, too, though.... might just take some more time.

leprechau:
That happens sometimes.  There is some odd race condition where vault doesn't launch before the test suite runs.  I need to spend some time debugging that but for now I'll just re-run that branch.  Thank you again!

valentin-krasontovitsch:
hmmm OK I see I've experienced such things myself, sure, thanks!

.... I'm not sure what vault is though, and I started looking at the first test that failed `TestTableLookup_656` and tried it with master locally myself, just running the subpackage `go test -count=1 ./route`, and it sometimes fails, sometimes don't - could that be due to a vault thing? it seems to my naive understanding to just be a simple redirect test... or is the vault the place where the table information is stored, and it happens async?

oh and cool thanks for merging : )

PR #495, opened 2018-05-16
Delete an unused global variable logOutput

gua-pian:
logOutput is not used as a global variable, it is re-initialized as short-term variable in func main().


magiconair:
Thank you!

Issue #494, opened 2018-05-16
Tests fail with consul > 1.0.6 and vault > 0.9.6


magiconair:
Consul startup times out and vault gets a 404. Need to investigate.

PR #497, opened 2018-05-17
Make tests compatible with Vault 0.10

pschultz:
When starting a new, empty Vault server, the mount at /secret will be
the K/V Version 2 backend by default. The API paths are subtly
different, but the new paths happen to also work with the Version 1
backend, so we can just always use the new paths in the test.

See https://www.vaultproject.io/docs/secrets/kv/kv-v2.html#acl-rules


pschultz:
With this patch all tests passed for me with Consul 1.0.6, 1.0.7, and 1.1.0, so this may fix #494 entirely.

magiconair:
hmm, why doesn't codeship pick this one up?

magiconair:
When I change the versions in `Makefile`

```patch
diff --git a/Makefile b/Makefile
index 6d2ff11..3a9ca94 100644
--- a/Makefile
+++ b/Makefile
@@ -27,8 +27,8 @@ GOVENDOR = $(shell which govendor)
 VENDORFMT = $(shell which vendorfmt)

 # pin versions for CI builds
-CI_CONSUL_VERSION=1.0.6
-CI_VAULT_VERSION=0.9.6
+CI_CONSUL_VERSION=1.1.0
+CI_VAULT_VERSION=0.10.1
 CI_GO_VERSION=1.10.2

 # all is the default target
```

 and run `make docker-test` I still get this:

```
--- FAIL: TestConsulSource (12.17s)
	source_test.go:240: Starting consul: Consul v1.1.0
	source_test.go:261: Timeout waiting for consul server after 12.0 seconds
=== RUN   TestVaultSource
--- FAIL: TestVaultSource (0.31s)
	source_test.go:305: Starting vault: "Vault v0.10.1 ('756fdc4587350daf1c65b93647b2cc31a6f119cd')\n"
	source_test.go:429: logical.Write failed: Error making API request.

		URL: PUT http://127.0.0.1:58421/v1/secret/data/fabio/cert/localhost
		Code: 400. Errors:

		* no data provided
```

pschultz:
Turns out the API isn't directly backward compatible after all. I amended my commit, so that Fabio detects a v2 backend and modifies Vault requests accordingly. It's not exactly pretty but it works. The new methods are almost verbatim copies from Vault's [cli implementation](https://github.com/hashicorp/vault/blob/master/command/kv_helpers.go).

This makes Fabio support the v2 backend automatically; no config change is required. The Vault policy must be updated, but that's not unique to Fabio. I seem to remember that a working sample policy is documented somewhere (in the wiki perhaps?) but I can't find it right now.

The Consul tests failed because your sneaky readiness check stopped working when they introduced gzip support, and I addressed that too.

PR #531, opened 2018-08-10
Add setting to flush fabio buffer regardless headers

samm-git:
## Description 

This PR adds ability to configure flush buffer interval for non-SSE request. 

In my configuration Fabio is serving services behind CDN. CDN expects any response to start (TTFB) within preconfigured timeout and if it is not - responds with 504. I can workaround this with setting SSE header, but in this case flush time is just 1s and also it requires to control every client which doing requests, and it is not always possible. This PR adds another setting, which can set this timeout globally, for all non-SSE requests. I am planning to set it to value which is high enough to not have significant performance impact but lower then cdn timeout.


magiconair:
LGTM. 

Could you also please add an entry to the website docs? https://github.com/fabiolb/fabio/tree/master/docs/content/ref

Thx

samm-git:
@magiconair hi, do you have any updates? Would be great to see it merged.

samm-git:
@magiconair ping again :)

samm-git:
@magiconair any chance to see it merged?

leprechau:
@samm-git Thank you for your patience.  I see the docs.  LGTM!

Thank you!

magiconair:
@samm-git thx for your patience

Issue #224, opened 2017-01-30
Routing table not updated when rule references non-existent service


rodriguezrps:
Fabio failed to create the entire route table because one of the manual override rule referenced to a nonexistent service. This was working on version 1.3.5 and stopped on version 1.3.6 and still failing on master. we are wondering if it's a bug or a feature? Our point of view is that a wrongly specified override shouldn't stop the entire table to be created.

magiconair:
The routing table is created from combining the generated and the manual rules by simply appending and then parsing them. Any parse error prevents the routing table from being updated. This keeps things simple and means less edge cases. However, this only refers to **syntax** errors. Referencing a nonexistent service is not a syntax error, though. So this sounds like a bug.

magiconair:
Hmm, I can't reproduce this on current master. I ran `consul agent -server -dev`, `./fabio` and `demo/server -prefix /foo` which registers a service `server` in consul. Then I've added the following rule to the manual overrides:

```
route add fump /bar http://1.2.3.4:5678/
```

which refers to a non-existent service on a non-existent endpoint. fabio picks up the routing table properly and calls to `localhost:9999/foo` still work as expected. 

Can you share your manual override entry which caused the error? Was there anything in the logs that could help track down the root cause?

magiconair:
Same behavior with 1.3.5, 1.3.6 and 1.3.7.

rodriguezrps:
yes, sure! I will do soon as I have time, probably later this evening.

magiconair:
@rodriguezrps Any update?

magiconair:
I'm going to close this for now. Please feel free to comment and update once you have more info and I'll re-open as necessary.

rodriguezrps:
Hi, sorry for the late reply, I was quite busy.
Well, I just simulated on my local within the following steps:
- Started a local consul and fabio
- started 2 demo app's (svc-a and svc-b)
- added 2 manual override's 
   route weight svc-a /foo weight 0.8
   route weight svc-c /bar weight 1.0
the weight of svc-a didn't change and I got '[WARN] route: no target match'
not sure if this should be the normal behavior, but as I said, it was working on version 1.3.5

magiconair:
@rodriguezrps I can reproduce this now. Might be a good idea to have an issue template which asks for steps to reproduce the problem, e.g.

```
# start consul
$ consul agent -server -dev

# start fabio
$ fabio

# start first server
[...]

PR #541, opened 2018-09-03
add support for 'weight_tags=<t1>,<t2>,… ’ option on the url tag

adesaegher:
This patch add support to _'weight_tags=\<t1>,\<t2>,...’_ option on the url tag which permits to configure weight by tags on a service as per documentation:   _route weight \<src> weight \<w> tags "\<t1>,\<t2>,… "_
First use case would be an active/passive service. With a tag like _"urlprefix-vault.example.io/ weight=1 weight_tags=active"_ , Fabio would always route to the active node.

### DIFF
--- a/registry/consul/service.go
+++ b/registry/consul/service.go
@@ -119,6 +119,7 @@ func serviceConfig(client *api.Client, name string, passing map[string]bool, tag
 
 				// build route command
 				weight := ""
+				weight_tags := ""
 				ropts := []string{}
 				tags := strings.Join(svctags, ",")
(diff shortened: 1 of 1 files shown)


adesaegher:
I got this error after a while:
2018/10/04 15:03:49 [WARN] route: no target match
It seems related to this issue : https://github.com/fabiolb/fabio/issues/224

leprechau:
I believe the intention here should already be possible with the current tags.  If this is incorrect please feel free to reopen.

PR #553, opened 2018-09-20
Added new case insensitive matcher

herbrandson:
This adds the ability to have case insensitive matching for routes. This is important for projects that need to migrate from other load balancers that are case insensitive.

### DIFF
--- a/config/load.go
+++ b/config/load.go
@@ -247,7 +247,7 @@ func load(cmdline, environ, envprefix []string, props *properties.Properties) (c
 		return nil, fmt.Errorf("invalid proxy.strategy: %s", cfg.Proxy.Strategy)
 	}
 
-	if cfg.Proxy.Matcher != "prefix" && cfg.Proxy.Matcher != "glob" && cfg.Proxy.Matcher != "noCase" {
+	if cfg.Proxy.Matcher != "prefix" && cfg.Proxy.Matcher != "glob" && cfg.Proxy.Matcher != "nocase" {
 		return nil, fmt.Errorf("invalid proxy.matcher: %s", cfg.Proxy.Matcher)
 	}
 
--- a/config/load.go
+++ b/config/load.go
@@ -247,7 +247,7 @@ func load(cmdline, environ, envprefix []string, props *properties.Properties) (c
 		return nil, fmt.Errorf("invalid proxy.strategy: %s", cfg.Proxy.Strategy)
 	}
 
-	if cfg.Proxy.Matcher != "prefix" && cfg.Proxy.Matcher != "glob" {
+	if cfg.Proxy.Matcher != "prefix" && cfg.Proxy.Matcher != "glob" && cfg.Proxy.Matcher != "noCase" {
 		return nil, fmt.Errorf("invalid proxy.matcher: %s", cfg.Proxy.Matcher)
 	}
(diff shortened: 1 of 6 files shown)


route/matcher.go
```diff
@@ -12,6 +12,7 @@ type matcher func(uri string, r *Route) bool
 var Matcher = map[string]matcher{
 	"prefix": prefixMatcher,
 	"glob":   globMatcher,
+	"noCase": noCaseMatcher,
```
herbrandson: I feel like this might be a terrible name, but it's the best I could come up with this evening. I'm open to suggestions

herbrandson:
I've never done GoLang before, but this seemed pretty straight forward based on the other examples. Let me know if I missed anything.

route/matcher.go
magiconair: Maybe `nocase`

route/matcher.go
```diff
@@ -23,3 +24,10 @@ func prefixMatcher(uri string, r *Route) bool {
 func globMatcher(uri string, r *Route) bool {
 	return r.Glob.Match(uri)
 }
+
+// noCase matches path to the routes' path ignoring case
+func noCaseMatcher(uri string, r *Route) bool {
+	lowerURI := strings.ToLower(uri)
+	lowerPath := strings.ToLower(r.Path)
+	return strings.HasPrefix(lowerURI, lowerPath)
```
herbrandson: After looking at this again, I think this should be a case insensitive _prefix_ match. Originally I was using `EqualFold`. Thoughts?

config/load.go
```diff
@@ -247,7 +247,7 @@ func load(cmdline, environ, envprefix []string, props *properties.Properties) (c
 		return nil, fmt.Errorf("invalid proxy.strategy: %s", cfg.Proxy.Strategy)
 	}
 
-	if cfg.Proxy.Matcher != "prefix" && cfg.Proxy.Matcher != "glob" && cfg.Proxy.Matcher != "noCase" {
+	if cfg.Proxy.Matcher != "prefix" && cfg.Proxy.Matcher != "glob" && cfg.Proxy.Matcher != "nocase" {
```
magiconair: With respect to the case-insensitive prefix matching I think `iprefix` is a better name.

fabio.properties
```diff
@@ -297,6 +297,7 @@
 #
 # prefix: prefix matching
 # glob:  glob matching
+# nocase: matches using a case insensitive test
```
magiconair:
```
# iprefix: case-insensitive prefix matching
```

route/matcher.go
magiconair:
```go
// todo(fs): if this turns out to be a performance issue we should cache
// todo(fs): strings.ToLower(r.Path) in r.PathLower
return strings.HasPrefix(strings.ToLower(uri), strings.ToLower(r.Path))
```

herbrandson:
I believe I've addressed all of your comments. I just wanted to make sure there wasn't something I missed that you're still waiting on. Are there any other changes you'd like to see on this?

herbrandson:
Any update on this? We really need this in order to migrate to fabio

magiconair:
Thank you and sorry that it took so long!

Issue #554, opened 2018-09-20
Enhance Glob matching by using compiled Globs


galen0624:
Glob Matching today has a performance hit when Fabio has a lot of services (>500).  This can impact CPU performance significantly.  As described in the `performance` section of the vendored package.
https://github.com/gobwas/glob

I have most of the code ready for using compiled Globs.  I will PR once I get through some of the load testing.

galen0624:
It doesn't appear that the compiled globs helped the CPU issue all that much.  My testing details are below.    I'm going to run a CPU Profile at high load and post the details. 
Load test on an 8core 8G VM.  

Sep 20 21:46:49  fabio: "GlobMatchingDisabled": true,
Sep 20 21:46:49  fabio: "GlobCacheSize": 5000

Test 1 - 10K TPS 37% CPU

Sep 20 21:51:26 fabio: "GlobMatchingDisabled": false,
Sep 20 21:51:26 fabio: "GlobCacheSize": 5000

Test 2 - 2K TPS 30% CPU
Test 3 - 3K TPS 38% CPU
Test 4 - 4K TPS 98% CPU (TPS didn't exceed 3.7K)
Test 5 - 10K TPS 98% CPU (TPS didn't exceed 4.0K)


<img width="1789" alt="4k-tps-with-without-globmatching" src="https://user-images.githubusercontent.com/4601212/45858067-41c80c80-bd21-11e8-935f-08d2a6cc9663.png">

galen0624:
We have about 1500 Services in this ENV.

glob.cache.size = 5000
profile.mode = cpu
profile.path = /tmp
5K TPS Test

Based on the CPU profile it looks like the RLock and RUnlock are taking most of the CPU.  I'm not sure how to avoid using those without creating race conditions and panics.  
Another func we could look at is the stings.ToLower but that is tertiary to the Mutex locks


[cpu1.pdf](https://github.com/fabiolb/fabio/files/2403820/cpu1.pdf)

galen0624:
I will PR my current code to double check my logic.

galen0624:
Is this something we could look into - https://golang.org/doc/go1.9#sync-map

galen0624:
If we moved the compiled glob cache add/remove functions to the Route add/del functions using the Mutex locks we in theory should be able to remove the fast path lookup mutex Rlock/UNlock.     That way on we move the CPU intensive part (Rlock/RUnlock) to the "control plane" side of Fabio instead of the "data plane" . 

[...]

PR #555, opened 2018-09-21
Added compiled glob matching using LRU cache

galen0624:
#554  Review for CPU usage of Mutex.


galen0624:
#554 The use of sync.Map increased the performance from 3500 TPS to 6500 TPS.  
The glob_cache_test file needs refactoring as you have to use sync.Map.Range and there is no ability to get the len of a sync.Map currently

galen0624:
#554 Updated glob_cache to support all types that implement the glob.Glob Interfaces.  Make test now passes all checks.

galen0624:
#554  Updated to remove type switch and assert type to glob.Glob interface.

magiconair:
This looks OK. If you rebase this then we can merge this as well. I assume you're using that in production as well?

galen0624:
We are not using this in prod.  We don't use GLOB matching.  That being said we have tested it.  -- https://github.com/fabiolb/fabio/issues/554#issuecomment-425276410

I will rebase line and PR.

galen0624:
rebased this PR.

galen0624:
Closing and recreating with new PR #615

Issue #558, opened 2018-10-11
Route updates are delayed with large number of services 


galen0624:
We have a large number of services > 2500.  The watchBackend func is taking over 15-20 seconds due to the sequential service queries.   I am proposing an update to the below code to be a go func with worker pools.

https://github.com/fabiolb/fabio/blob/master/registry/consul/service.go#L57

I will PR this hopefully early next week.

galen0624:
Closing with #564

magiconair:
I'll keep this open as the issue which is the motivation for the PR.

PR #559, opened 2018-10-23
Issue 558 concurrent service updates

galen0624:
#558

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -145,4 +145,5 @@ type Consul struct {
 	CheckTLSSkipVerify                  bool
 	CheckDeregisterCriticalServiceAfter string
 	ChecksRequired                      string
+	ConcurrentConsulRequests            int
 }
--- a/config/default.go
+++ b/config/default.go
@@ -62,6 +62,7 @@ var defaultConfig = &Config{
 			CheckScheme:                         "http",
 			CheckDeregisterCriticalServiceAfter: "90m",
 			ChecksRequired:                      "one",
+			ConcurrentConsulRequests:            10,
 		},
 		Timeout: 10 * time.Second,
 		Retry:   500 * time.Millisecond,
--- a/config/load.go
+++ b/config/load.go
@@ -169,6 +169,7 @@ func load(cmdline, environ, envprefix []string, props *properties.Properties) (c
 	f.StringVar(&cfg.Registry.Consul.KVPath, "registry.consul.kvpath", defaultConfig.Registry.Consul.KVPath, "consul KV path for manual overrides")
 	f.StringVar(&cfg.Registry.Consul.NoRouteHTMLPath, "registry.consul.noroutehtmlpath", defaultConfig.Registry.Consul.NoRouteHTMLPath, "consul KV path for HTML returned when no route is found")
 	f.StringVar(&cfg.Registry.Consul.TagPrefix, "registry.consul.tagprefix", defaultConfig.Registry.Consul.TagPrefix, "prefix for consul tags")
+	f.IntVar(&cfg.Registry.Consul.ConcurrentConsulRequests, "registry.consul.concurrentrequests", defaultConfig.Registry.Consul.ConcurrentConsulRequests, "number of concurrent requests querying consul")
 	f.BoolVar(&cfg.Registry.Consul.Register, "registry.consul.register.enabled", defaultConfig.Registry.Consul.Register, "register fabio in consul")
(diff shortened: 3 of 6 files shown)


magiconair:
Hi @galen0624, thanks for the patch! After looking at it I thought that this was an excellent opportunity to finally refactor the service monitor code for Consul and also add some tests for the route generation code. I hope that is OK with you?

I've also taken a slightly different approach for the concurrent fetching by limiting the number of concurrent go routines running. 

It would be awesome if you could test #564 and see whether it has the desired effect.

One thing I've noticed is that the current code also looks up the datacenter for every service. Moving this out of the code should already save some cycles but generally that is a cheap call.

I am curious on why you a) created a custom http client for the consul client and b) why you limited the number of idle connections. Could you elaborate on this?

magiconair:
I'll close this in the meantime.

magiconair:
#564 supersedes this.

magiconair:
Hmm, should this then be in the common code? Does it hurt?

galen0624:
No it shouldn't.  I apologize I neglected to remove it before the PR.

magiconair:
No need to apologize. :)

galen0624:
@magiconair  OK I tested the update and it behaved as expected.  I observed a 5-10% CPU increase using `registry.consul.serviceMonitors = 100` on an 8 Core box.  Memory was not impacted that I could tell.   

 I was also thinking about the issue with the custom http.Transport.  If the consul backend is consistent (meaning same URL/Host) the TIME_WAIT issue will still be something to solve regardless if Fabio is talking to localhost or a remote consul.  The `MaxIdleConnsPerHost` is per host.  During out testing today I observed around 25K TIME_WAIT connections with roughly 2000 services. 

 Below is somewhat of an explanation on the settings.  

http://tleyden.github.io/blog/2016/11/21/tuning-the-go-http-client-library-for-load-testing/

galen0624:
With 2000 + services there is always changes.  The Consul blocking query is irrelevant in our ENV because of that.   I instituted a configurable query timer in the "Shim" instead.  It acts like a blocking query to Fabio.  The expectation is that route updates in Fabio happen within the query timer interval.  Without concurrent service updates each Fabio was taking 20+ seconds to run through all of the services.   Additionally we have sized our Consul ENV to support the amount of queries that we are sending from the FabioShims.  

I'll continue testinging with various `registry.consul.serviceMonitors` and see where it breaks down.

galen0624:
After testing it seems that `registry.consul.serviceMonitors = 16` is the sweet spot for meeting our current query interval.

galen0624:
Also I have updated and tested your issue-558 branch with the new custom http.Transport if you want me to PR it back.

leprechau:
To the earlier point of the number of Fabio instances and service monitors and the potential load on Consul ... would it make sense to incorporate the ability to use the new agent cache added upstream in v1.3.0?

leprechau:
Maybe make an available global tunable to set a cache interval for the health calls?  I could introduce a separate PR for that specifically.

magiconair:
I was thinking the same thing. Either wait for changes or poll. I think this would go nice with the refactoring. I could add @galen0624 as contributor so that we can work on the same branch instead of merging small bits back and forth.

magiconair:
Just did. @galen0624 do you have a GPG key to sign your commits? I should update the contrib guidelines with that.

magiconair:
If not, here is a good guide on how to create a good set of keys: https://github.com/drduh/YubiKey-Guide

magiconair:
#564 is on the `issue-558` branch

galen0624:
I have one through KeyBase.  Will that work?  If not I'll get another one setup tonight.

galen0624:
@leprechau Can you elaborate on the V1.3.0?  Is that Consul?  

Are you talking about the Agent Cache described below for /catalog/service/my-serivce

https://www.consul.io/api/catalog.html

galen0624:
@magiconair @leprechau I stared the updates for polling intervals.  Not 100% complete yet but a starting point.  I'm not sure if you want me to submit a PR or commit to the issue-558 branch?  The code can be found here for now.

https://github.com/galen0624/fabio/tree/issue-558

magiconair:
@galen0624 I've left some comments on https://github.com/galen0624/fabio/commit/474d896768c761d68ce60142fafaea6703eb9616. 

Any objection on merging #564?

galen0624:
@magiconair No objections on Merging #564.  We have been using it in production for the last few weeks.  I will be updating the other request and submitting an update based on your feedback.

galen0624:
I will.

Issue #562, opened 2018-10-26
Change Log entry update 


galen0624:
Update Change log `glob.matching.enabled` to be `glob.matching.disabled`

PR #563, opened 2018-10-26
issue 562 update change log glob.matching.disabled

galen0624:
#562 
Updated change log `glob.matching.enabled` to be `glob.matching.disabled`


magiconair:
Thx!

PR #564, opened 2018-10-27
Refactor consul service monitor

### LINKED ISSUE #558: Route updates are delayed with large number of services

magiconair:
This patch refactors the consul service monitor and provides a new option `registry.consul.serviceMonitors` to configure the concurrency for routing table updates. 

Fabio uses one go routine by default to fetch status information from Consul when the service status changes. This can be too slow when lots of services are registered. The new `registry.consul.serviceMonitors` option configures how many go routines fabio will use to fetch service data from Consul.

This patch unfortunately supersedes #559 from @galen0624 since this was a good opportunity to finally refactor the service monitor code to make it more extensible and testable.

Fixes #558


magiconair:
Discussion is on #559

Issue #166, opened 2016-09-27
Basic authentication on routes


ygersie:
In Operations we often have to deal with Basic auth protected sites. The fabio-ui (#25) itself or announced routes for Nomad, Consul UI etc should be protected. The way we have to deal with that right now is setting up separate clusters consisting of:
- Public fabio
- Internal fabio (non protected)
- Internal fabio (ldap protected behind a proxy)

It would be (more than) nice to be able to announce in the route itself that the resource should be protected with auth-type=ldap to perform Basic Auth. This saves us from a lot of configuration headaches and wasting resources on separating fabio instances and proxies and creating more VIPs to point to the correct clusters.

tino:
Or how about fetching the basic auth credentials from Vault? Seems like a sensible place to put 'em.

csawyerYumaed:
or authenticate against vault auth directly. Then you can get LDAP or user/pass/etc auth "for free" as it were, without having to bring in all the LDAP libs directly into fabio.  Bonus would be getting assigned vault policies to decide if they have access or not.

magiconair:
@csawyerYumaed That sounds like an interesting idea.

nmaludy:
+1 for some sort of auth on routes, i really like the vault integration idea!

leprechau:
So out of curiosity how do you see this working?

possible options ...

vault that would take a method and raw kv arguments...
```
opts "auth=vault:method=method:arg1=value:arg2=value"
```

http basic auth...
```
opts "auth=basic:user1:pass1,user1:pass2"
```

http digest auth...
```
opts "auth=digest:user1:realm1:hash1,user2:realm2:hash2"
```

leprechau:
The reason I'm asking ... I could see this being an expansion to the ACL support that was recently added.

nmaludy:
@leprechau Yeah i think something along the lines of:

```
# auth : the name of the auth provider in the config (for things like secrets and htpasswd paths)
# type : what type of auth backend this is (vault, basic, digest)
# path : this is vault specific, maybe it makes sense to have this in the config?
# arg1=value1 : Backend specific key/value pairs to set in the auth request API call (for Vault) or maybe username=myuser for basic auth
auth=myvault:type=vault:method=ldap:path=ldap:arg1=value1:arg2=value2
```

Thinking the Vault token should probably be stored in the Fabio config instead of the route tags.

Also probably makes sense for basic auth to store the local path of the htpasswd file in the Fabio config?

[...]

PR #573, opened 2018-11-16
add http-basic auth reading from a file

andyroyle:
Taking implementation cues from #166 I've made an implementation of http basic auth on a per-route basis.

The auth schemes need to be registered with fabio up-front using

`proxy.auth = name=mybasicauth;type=basic;file=path/to/file`

You then configure your route specifying `auth=mybasicauth` in the route options.

You can configure multiple auth schemes side-by-side

```
proxy.auth = name=mybasicauth;type=basic;file=path/to/file;realm=foo
                      name=myotherauth;type=basic;file=path/to/other/file
```

Currently it uses http basic auth, which reads a single htpasswd file at startup from a file on disk.

If you specify an auth-scheme that doesn't exist then it will log and error in fabio and return a `401 Unauthorized` response.

You can optionally set the realm (default is to use the auth name parameter)


leprechau:
Thoughts of supporting htpasswd format for the file?

andyroyle:
Dead easy to add, do we want to make htpasswd the default for basic auth?

leprechau:
I would lean to making that default.  The tooling to create htpasswd files is already present on most *nix systems and users of a load balancer are probably familiar with them and know how to create/manage those files.  What do you think?

Sideline: Thanks for picking up the torch on that old issue.

andyroyle:
Htpasswd is pretty standard stuff. Will make the changes first thing Monday :+1:

leprechau:
Awesome, and thank you again!

magiconair:
Thank you and yes to htpasswd.

andyroyle:
@magiconair  / @leprechau htpasswd support added. 

Also added support for an optional realm (used in the `WWW-Authenticate: Basic realm="<realm>"` response)

andyroyle:
So I vendored in a new library, which depends on `golang.org/x/crypt` but it seems to have brought with it updates to `golang.org/x/net` and `golang.org/x/text`, yet `go.mod` only shows a single change. Colour me confused.

shantanugadgil:
I stumbled upon this, and I wonder if it would solve my use case?
(Looking at the diff it seems so, based on the explanation, atleast)

My use case it to provide an "https + password" **frontend** to the Consul/Nomad GUI via Fabio.
Currently, I am making do with HAProxy http basic auth on a separate port.

Eagerly looking forward to this getting merged. :clap:  :clap:

andyroyle:
@shantanugadgil this is exactly what we are using it for as well 😄

leprechau:
@andyroyle Sorry for the delay, just getting caught back up after a short vacation.  I'll try to give this a review today.

auth/auth_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+func TestLoadAuthSchemes(t *testing.T) {
+
+	t.Run("should fail when auth scheme fails to load", func(t *testing.T) {
+		_, err := LoadAuthSchemes(map[string]config.AuthScheme{
+			"myauth": {
+				Name: "myauth",
+				Type: "basic",
+				Basic: config.BasicAuth{
+					File: "/some/non/existant/file",
```
magiconair: `existent` :)

auth/auth_test.go
```diff
@@ -0,0 +1,76 @@
...
+				},
+			},
+		})
+
+		const errorText = "open /some/non/existant/file: no such file or directory"
```
magiconair: same here...

auth/basic_test.go
```diff
@@ -0,0 +1,188 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	rw := &responseWriter{}
+
+	_ = basicAuth.Authorized(&http.Request{Header: http.Header{}}, rw)
+
+	got := rw.Header().Get("WWW-Authenticate")
+	want := "Basic realm=\"testrealm\""
```
magiconair: you can use string literals here, e.g. `` `Basic realm="testrealm"` ``

docs/content/feature/authorization.md
```diff
@@ -0,0 +1,54 @@
+---
+title: "Authorization"
+since: "1.5.10"
```
magiconair: This would need to be 1.5.11

magiconair:
@andyroyle  can you rebase this, please? Then I'll merge it.

andyroyle:
@magiconair thanks for the review (and for merging #575). I've fixed those issues and also rebased on top of master, so this should be good-to-go now.

PR #574, opened 2018-11-20
metrics.circonus: Add support for circonus.submissionurl

stack72:
There are times when nodes that fabio are deployed on already have a circonus metrics agent added to them. Following the pattern that Consul, Nomad and Vault have taken, we can actually send the application metrics to the local agent by using the SubmissionURL

I am currently using this in our environment with the config as follows:

```
metrics.target=circonus
metrics.circonus.submissionurl=http://127.0.0.1:2609/write/fabio
```


magiconair:
Thanks Paul!

PR #436, opened 2018-02-07
Add HTTP/2 support for gRPC

magiconair:
Fixes #244

PR #575, opened 2018-11-23
Add GRPC proxy support

andyroyle:
Adds proper GRPC proxy support, including TLS upstream and TLS termination.


magiconair:
Thanks a lot @andyroyle for this. Are you running this in prod already?

stack72:
@magiconair yeah we have this in production now for about 3 weeks - it was a replacement for us using HaProxy and it’s integrated well with consul :)

magiconair:
Thx Paul.

andyroyle:
So this PR doesn't include those http/2 changes from #436 in, because it creates an entirely separate transport. It might still be worth merging that.

PR #581, opened 2018-12-07
address concerns raised while troubleshooting #524

### LINKED ISSUE #524: TCP proxy broken since v1.5.8
**Impact** 
It's not possible to connect to MySQL since v1.5.8

**STR**
* Setup fabio as TCP proxy that listen 3306 port and has rule to route traffic to MySQL container (fabio version v1.5.7)
* Upgrade fabio to v1.5.8

**Expected result**
mysql client can establish connection to MySQL service

**Actual result**
mysql client hangs during connection establishment.

**Additional info**
v1.5.7
```
$ nc 10.0.0.10 3306
J
5.6.40G!
B\'J}[0z��9KM"^612FK%umysql_native_password
```

v1.5.8
```
$ nc 10.0.0.10 3306
<empty output, process hangs until enter press>
J
5.6.40G!
B\'J}[0z��9KM"^612FK%umysql_native_password
```

**Configuration**

*proxy_addr* 
```
:3306;proto=tcp,:80,:22;proto=tcp,:23;proto=tcp,:8080,:1025;proto=tcp,:5672;proto=tcp,:5673;proto=tcp;cs=certstore,:443;cs=certstore,:8125;proto=tcp,:9000;cs=certstore,:27017;proto=tcp
```

*mysql* service register at Consul with following tag: `urlprefix-:3306 proto=tcp`

leprechau:
This PR, when merged, should resolve #524 and restore proper TCP proxy functionality with an out-of-box configuration.

### DIFF
--- a/route/access_rules.go
+++ b/route/access_rules.go
@@ -68,7 +68,7 @@ func (t *Target) AccessDeniedHTTP(r *http.Request) bool {
 
 // AccessDeniedTCP checks rules on the target for TCP proxy routes.
 func (t *Target) AccessDeniedTCP(c net.Conn) bool {
-	// Calling RemoteAddr on a proxy-protocol enabled connection.
+	// Calling RemoteAddr on a proxy-protocol enabled connection may block.
 	// Therefore we explicitly check and bail out early if there are no
 	// rules defined for the target.
 	// See https://github.com/fabiolb/fabio/issues/524 for background.
--- a/route/access_rules.go
+++ b/route/access_rules.go
@@ -17,7 +17,7 @@ const (
 // AccessDeniedHTTP checks rules on the target for HTTP proxy routes.
 func (t *Target) AccessDeniedHTTP(r *http.Request) bool {
 	// No rules ... skip checks
-	if t.accessRules == nil || len(t.accessRules) == 0 {
+	if len(t.accessRules) == 0 {
 		return false
 	}
 
@@ -72,7 +72,7 @@ func (t *Target) AccessDeniedTCP(c net.Conn) bool {
 	// Therefore we explicitly check and bail out early if there are no
 	// rules defined for the target.
 	// See https://github.com/fabiolb/fabio/issues/524 for background.
-	if t.accessRules == nil || len(t.accessRules) == 0 {
+	if len(t.accessRules) == 0 {
 		return false
 	}
 	// validate remote address assertion
--- a/route/access_rules.go
+++ b/route/access_rules.go
@@ -16,16 +16,20 @@ const (
 
 // AccessDeniedHTTP checks rules on the target for HTTP proxy routes.
 func (t *Target) AccessDeniedHTTP(r *http.Request) bool {
-	var ip net.IP
-	host, _, err := net.SplitHostPort(r.RemoteAddr)
+	// No rules ... skip checks
+	if t.accessRules == nil || len(t.accessRules) == 0 {
+		return false
+	}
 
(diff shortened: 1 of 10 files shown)


leprechau:
@pschultz @magiconair PR to address #524

route/access_rules.go
```diff
@@ -16,16 +16,20 @@ const (
 
 // AccessDeniedHTTP checks rules on the target for HTTP proxy routes.
 func (t *Target) AccessDeniedHTTP(r *http.Request) bool {
-	var ip net.IP
-	host, _, err := net.SplitHostPort(r.RemoteAddr)
+	// No rules ... skip checks
+	if t.accessRules == nil || len(t.accessRules) == 0 {
```
magiconair: The `nil` check is redundant.

route/access_rules.go
```diff
@@ -64,16 +68,21 @@ func (t *Target) AccessDeniedHTTP(r *http.Request) bool {
 
 // AccessDeniedTCP checks rules on the target for TCP proxy routes.
 func (t *Target) AccessDeniedTCP(c net.Conn) bool {
-	var addr *net.TCPAddr
-	var ok bool
+	// Calling RemoteAddr on a proxy-protocol enabled connection may block.
+	// Therefore we explicitly check and bail out early if there are no
+	// rules defined for the target.
+	// See https://github.com/fabiolb/fabio/issues/524 for background.
+	if t.accessRules == nil || len(t.accessRules) == 0 {
```
magiconair: same here

magiconair:
Is it possible to have multiple authors on a single commit? Then you could attribute @pschultz as well. More curious than asking.

leprechau:
I'm not sure about the commit itself ... I don't know.  I fully agree though that @pschultz should get credit in the changelog and release notes at a minimum.

route/access_rules.go
leprechau: Good point.  I always forget that nil and typed nil behave very differently.

leprechau:
@magiconair Any more thoughts on changing the default state of proxy-protocol on the listener?  Do we want to keep it false as-is in the current PR?

I also noticed that in the updated `go-proxyproto` package there is a function to check for allowed sources for proxy-protocol.  That would elevate some of the security concerns but I think that's a later feature enhancement.

magiconair:
You'll learn something new every day. Thx for looking.

sean-:
@magiconair, no it wouldn't.  We're not using the TCP `PROXY` protocol anywhere.  We're using gRPC now (successfully) and auth.

magiconair:
Lets do this. Thank you all!

magiconair:
@leprechau shoot, we forgot something. Can you provide another PR with an updated `fabio.properties` and documentation? Feel free to merge right away. I'll work on the CHANGELOG.

leprechau:
@magiconair absolutely.

magiconair:
Also, fabio users: if you find this change and think disabling PROXY by default was a grave mistake please let us know. Thx

magiconair:
@leprechau does this disable proxy for tcp only or also for http?

leprechau:
This disables PROXY protocol on all listeners by default unless enabled with `pxyproto=true`.

magiconair:
I'm having second thoughts on disabling by default. Will sleep on it a night. :)

leprechau:
Okay, I have the documentation done.  I'll push it to a new branch and create a PR.

leprechau:
Updated documentation here: https://github.com/fabiolb/fabio/pull/583

PR #582, opened 2018-12-07
be consistent with checks in access rules

leprechau:


leprechau:
oops... that branch name no longer applies

PR #583, opened 2018-12-07
update documentation around the changes to PROXY protocol

leprechau:


leprechau:
Merging this to ensure master is functional and documentation matches current master functionality.  We can update/revert as needed if we decide to change direction at a later date.

Issue #508, opened 2018-06-13
could the source and destination be clickable in the ui?


kneufeld:
First of all, thanks so much for fabio, it's awesome and I love it.

Minor feature request though, could you make the source and especially the destination columns in the ui an href so that you can go directly to that service in the browser? If the service has an http check (as opposed to a tcp check) then service name could also be a link.

Thanks!

magiconair:
Thank you!

You can submit a PR since this looks like a simple change. All the code is in https://github.com/fabiolb/fabio/blob/master/admin/ui/route.go

Happy to merge it.

magiconair:
Closed by #587

PR #587, opened 2018-12-20
make Dest column into clickable links

kneufeld:
addresses my issue #508

### DIFF
--- a/admin/ui/route.go
+++ b/admin/ui/route.go
@@ -91,7 +91,7 @@ $(function(){
 			tbl += '<td>' + (i+1) + '</td>';
 			tbl += '<td>' + r.service + '</td>';
 			tbl += '<td>' + r.src + '</td>';
-			tbl += '<td>' + r.dst + '</td>';
+			tbl += '<td><a href="' + r.dst + '">' + r.dst + '</a></td>';
 			tbl += '<td>' + r.opts + '</td>';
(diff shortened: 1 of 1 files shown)


admin/ui/route.go
```diff
@@ -91,7 +91,7 @@ $(function(){
 			tbl += '<td>' + (i+1) + '</td>';
 			tbl += '<td>' + r.service + '</td>';
 			tbl += '<td>' + r.src + '</td>';
-			tbl += '<td>' + r.dst + '</td>';
+			tbl += '<td><a href="' + r.dst + '">' + r.dst + '</a></td>';
```
pschultz: Does the attribute have to be escaped? What happens if r.dst contains a quote, for instance?.

pschultz: I suppose this code is already vulnerable to XSS. I'm gonna fix this in another PR.

kneufeld: I've only seen `dst` be an ip:port combo, not sure how clicking on a link has anything to do with XSS?

pschultz: `dst` is an arbitrary string from the configuration and may contain `<script>evilCode();</script>`. But again, I've addressed this in #588, so don't worry about it here.

magiconair:
Resolved this and I'll merge this for 1.5.11. Hope I did this correctly. @pschultz could you have a quick look, please?

pschultz:
@magiconair, yes, looks good to me.

PR #391, opened 2017-11-20
 Fabio can connect to Consul using TLS for secure communication.

jrasell:
This change updates Fabio's configuration parameters to allow use and specification of TLS certificates for HTTPS. In Consul setups where full TLS is configured using CA private this allows Fabio to work in a secure manner.

I have tested this change locally with a Consul TLS setup; log fragments are shown below to provide further detail:

```
./fabio -registry.consul.caFile=/Users/rasellj/opt/hashicorp/ca/rasellj-ca.pem -registry.consul.enableSSL -registry.consul.keyFile=/Users/rasellj/opt/hashicorp/consul/consul-server-key.pem -registry.consul.certFile=/Users/rasellj/opt/hashicorp/consul/consul-server.pem -registry.consul.verifySSL
```

```
"EnableSSL": true,
"VerifySSL": true,
"CAFile": "/Users/rasellj/opt/hashicorp/ca/rasellj-ca.pem",
"CertFile": "/Users/rasellj/opt/hashicorp/consul/consul-server.pem",
"KeyFile": "/Users/rasellj/opt/hashicorp/consul/consul-server-key.pem"
```

```
2017/11/20 09:44:50 [INFO] Version v1.5.3-10-g5b2fe34 starting
2017/11/20 09:44:50 [INFO] Go runtime is go1.9
2017/11/20 09:44:50 [INFO] Metrics disabled
2017/11/20 09:44:50 [INFO] Setting GOGC=800
2017/11/20 09:44:50 [INFO] Setting GOMAXPROCS=8
2017/11/20 09:44:50 [INFO] consul: Connecting to "localhost:8500" in datacenter "dc1"
2017/11/20 09:44:50 [INFO] Admin server access mode "rw"
2017/11/20 09:44:50 [INFO] Admin server listening on ":9998"
2017/11/20 09:44:50 [INFO] Waiting for first routing table
2017/11/20 09:44:50 [INFO] consul: Using dynamic routes
2017/11/20 09:44:50 [INFO] consul: Using tag prefix "urlprefix-"
2017/11/20 09:44:50 [INFO] consul: Watching KV path "/fabio/config"
2017/11/20 09:44:53 [INFO] consul: Registered fabio with id "REDACTED-9998"
2017/11/20 09:44:53 [INFO] consul: Registered fabio with address "10.210.90.45"
2017/11/20 09:44:53 [INFO] consul: Registered fabio with tags ""
[...]

PR #602, opened 2019-02-09
added support for Consul TLS transport

### LINKED ISSUE #276: Support Consul TLS
Currently only specifying the address is allowed, but there are many other options for Consul TLS like the CA file, key, and cert.

sev3ryn:
This change adds option for Fabio to connect Consul using TLS transport.

Copy of PR #391 while it seems to be incorrect and abandoned.

Closes #276


pires:
#391 can then be closed, no?

PR #606, opened 2019-03-01
docs: fix layout without JS enabled

### LINKED ISSUE #599: Sidebar of the website is a little off
https://fabiolb.net/

"Features" is partly hidden on Firefox.

image

pschultz:
The Bootstrap stylesheet for the website sets the header to `position:
fixed`, taking it out of the layout flow. Some javascript later updates
the style attribute, so that the content is pushed down by an
appropriate amount.

If JS is disabled that correction doesn't happen, and so the header
overlaps the content.

Make the header static by default, and only apply `position: fixed` if
JavaScript is enabled.

Fixes #599

### DIFF
--- a/docs/static/css/custom.css
+++ b/docs/static/css/custom.css
@@ -3,6 +3,12 @@
 	height: 100%;
 }
 
+/* set navbar to static for now. If JS is enabled, we will revert to fixed in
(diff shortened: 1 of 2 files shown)


magiconair:
Thank you!

PR #617, opened 2019-03-10
Fix newBasicAuth function to pass the test

mfuterko:
Hi! @galen0624, this may fix the build issue.


galen0624:
Looks good.  Thank you.

Issue #613, opened 2019-03-08
Adding multiple TCP listeners from Consul


murphymj25:
We have been working on providing a solution for TCP load balancing for a potential large number of applications.  Our requirements are that the service can be defined in consul, and the load balancer will automatically start the TCP listener.  Also, because we could have different applications requiring the same TCP port, we need to define the route by IP address.

With the current TCP implementation in Fabio, I am only able to start a listener in the properties file, and define a backend route using :port which doesn't match our current use case.

We have built a new proxy option that we are currently calling "tcp-dynamic" that uses the consul urlprefix tagging to start the TCP listeners and defines the backend route using ip:port.  Our initial testing has been promising, we were able to define 5000 different TCP ports in consul and start the listeners with pretty minimal CPU increases.  We need to do a little more cleanup, but should have something ready for feedback early next week.

murphymj25:
@craighazen and I have created pull request #626 for this issue.  Take a look and let us know if there is any feedback.

scalp42:
I'm currently trying to do the same thing using consul-template but I'm not getting very far.

Pseudo code would be:

- iterate over every service passing
- check the tags include "proto=tcp"
- update the `proxy.addr` in `fabio.properties` to include the `urlprefix-` tag and do some cleanup

Something like this (pseudo template):

```go
{{ range services }}
{{ if .Tags | join "," | regexMatch "proto=tcp" }}
<< something here to extract `fabio-:10000 proto=tcp` >> 
proxy.addr = :9999,<<:10000;proto=tcp>>
{{ end }}
{{ end }}
```

It'd be much better to just have this dynamic for sure.

PR #626, opened 2019-03-21
Issue 613 tcp dynamic

murphymj25:


murphymj25:
@craighazen and I have put together this PR to address #613.  This allows you to start a "tcp-dynamic" proxy that will start TCP listeners based on the consul urlprefix.  It will also match the service based on ip:port.  Please review and let us know if there is any feedback.

config/load.go
```diff
@@ -409,7 +415,7 @@ func parseListen(cfg map[string]string, cs map[string]CertSource, readTimeout, w
 	if l.Addr == "" {
 		return Listen{}, fmt.Errorf("need listening host:port")
 	}
-	if csName != "" && l.Proto != "https" && l.Proto != "tcp" && l.Proto != "grpcs" {
+	if csName != "" && l.Proto != "https" && l.Proto != "tcp" && l.Proto != "tcp-dynamic" && l.Proto != "grpcs" {
 		return Listen{}, fmt.Errorf("cert source requires proto 'https', 'tcp' or 'grpcs'")
```
leprechau: update the error message to match the conditional

murphymj25: Good catch, i'll get that updated.
Thanks!

murphymj25: @leprechau, I have updated the error message to match the conditional.

Thanks!

Michael

murphymj25:
I need to do some more updates to this.  Manual overrides cause a panic.

scalp42:
@murphymj25 thanks a lot for the contribution. It fixed the issue of having to define the TCP in advance and allow the tags to be the authoritative source.

murphymj25:
@scalp42 Let me know how your experience goes.  We're just starting to get a few services deployed using this, so probably a few things we'll need to update yet.

leprechau:
@murphymj25, @scalp42 Yes, absolutely.  Let us know how the testing goes and we'll look at getting this merged.

scalp42:
I think we're at the crossroad of leveraging consul-template see #613, running custom Fabio binary (which we really don't want to) or just leaving the dynamic nature behind regarding Fabio config file (probably where we're headed).

murphymj25:
I found one error where we had the error catch for the dialer after the defer to close the connection.  This resulted in a panic if a connection could not be established.   I updated the PR with the correct order.

murphymj25:
Other than this one error, we have several apps using this, everything has been working well.

madsholden:
Any chance of this being merged and released soon? It would be really nice to have.

leprechau:
@madsholden I'll take a review today.  Thank you all for your patience.  @murphymj25 Thank you for all the testing and updates!

murphymj25:
No problem!  We haven't encountered any other issues apart from the one update I made.  We probably have around 20-30 different services running via this manner.

leprechau:
> No problem! We haven't encountered any other issues apart from the one update I made. We probably have around 20-30 different services running via this manner.

Awesome, thanks again.

sunliusi:
When will the new version be released? This is a great feature

renesnezic:
You mentioned this will match based on ip:port.. Does it match based on host:port?

I've got a scenario where I'm running multiple mail servers inside of separate docker containers..
Meaning I'd like to route traffic from example.com on ports 587 & 143 to one container and traffic for acme.com on ports 587 & 143 to a different container. And I obviously don't want to hard-code the container private IPs. Would this work?

PR #631, opened 2019-03-26
Minimize load on busy Consul

DukeyToo:
The current Fabio strategy for watching Consul services is to monitor globally for any health check changes, then query the Consul Catalog once for every passing service in Consul, then update the routing table with the changes. With enough Fabio instances, this strategy can generate significant load on larger Consul installations, due to the volume of health changes, and the large number of services that are each checked individually. 

In our specific case, we have many Fabio instances, but are routing only a few services in Consul.  Many changes in health are not relevant to our Fabio routing. This PR makes 2 adjustments in order to lighten the load:

1) Only query the catalog if the service has a matching tag prefix
2) Only query the catalog when something changed (many health check changes do not alter the status of any of the watched services).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ fabio.sublime-*
 demo/cert/
 /pkg/
 dist/
+custom.properties
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@
 .DS_Store
 .idea
 .vagrant
+.vscode
 build/builds/
 fabio
 fabio.exe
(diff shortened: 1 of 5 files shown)


DukeyToo:
I'm not sure if I should wrap the new behavior (for only checking valid tag prefix) in a new config flag. 

I imagine there may be some Fabio installations where they only use the dynamic KV routes, and some monitored consul services may not have any routing tags?

leprechau:
@DukeyToo Sorry, this was overlooked.  Were you able to resolve the issue or did you just give up?

DukeyToo:
The PR closed by itself because I deleted the branch on my fork - wasn't intentional. I'm not going to the trouble of reopening it though.

PR #632, opened 2019-03-29
Issue 605 grpc host matching

tommyalatalo:
We have implemented grpc host matching by using client-side metadata which is used to match hosts. If the specified metadata host is not found, the request will be routed according to the current default behavior, i.e. to any host serving the grpc service requested.

In doing this we also found that the GRPC max message size is hardcoded in fabio to the default 4mb, we increased this to 1Gb.


mwalters-workmarket:
This would be very useful for me.
Any timeline on getting merged?

leprechau:
Triggering a rebuild ... will merge if tests come back clean.  We have Travis setup again.

leprechau:
@tommyalatalo tests pass and I've looked over the code.  Would you be willing to add some documentation updates to describe the metadata options?

ishworg:
Would be good to have the metadata doco updated to get a better understanding.

cheers

rocwenlinux:
doc is updated, please check if that is enough.

docs/content/feature/grpc-proxy.md
```diff
@@ -14,6 +14,54 @@ fabio -proxy.addr ':1234;proto=grpc'
...
+```
+	md := metadata.Pairs("dsthost", "testonly")
+	ctx := metadata.NewOutgoingContext(context.Background(), md)
+```
+To make RPC using the context with the metadata will try to access the service with tag
+```
+ urlprefix-testonly/my.service/Method proto=grpc
+```
+
+first, if no such service exits, fall back to the service with tag
```
ishworg, line 29: Perhaps rephrase it as: <strikethrough>first,</strikethrough> If no service *exists*, fallback ...

docs/content/feature/grpc-proxy.md
```diff
@@ -14,6 +14,54 @@ fabio -proxy.addr ':1234;proto=grpc'
 
 As per the HTTP/2 spec, the host header is not required, so host matching is not supported for GRPC proxying.
 
+To route to a particular grpc service(s), set the 'dsthost'='name' in metadata, the grpc traffic will be routed to the service advertising `urlprefix-name/my.service/Method proto=grpc`.
+
+For example in Go grpc request:
```
ishworg, line 19: Perhaps `gRPC` instead of `grpc`?

docs/content/feature/grpc-proxy.md
```diff
@@ -14,6 +14,54 @@ fabio -proxy.addr ':1234;proto=grpc'
 
 As per the HTTP/2 spec, the host header is not required, so host matching is not supported for GRPC proxying.
 
+To route to a particular grpc service(s), set the 'dsthost'='name' in metadata, the grpc traffic will be routed to the service advertising `urlprefix-name/my.service/Method proto=grpc`.
```
ishworg, line 17:
```suggestion
To route to a particular grpc service(s), set the 'dsthost'='name' in metadata, the gRPC traffic will be routed to the service advertising `urlprefix-name/my.service/Method proto=grpc`.
```

proxy/grpc_handler.go
```diff
@@ -170,7 +170,6 @@ func (g GrpcProxyInterceptor) lookup(ctx context.Context, fullMethodName string)
 func (g GrpcProxyInterceptor) getDestinationHostFromMetadata(md metadata.MD) (dstHost string) {
 	dstHost = ""
 	hosts := md["dsthost"]
-	fmt.Println("hosts", md)
 	if len(hosts) == 1 {
 		dstHost = hosts[0]
 	}
```
ishworg, line 175: I know it is outside the scope of the review right now but this can be more concisely written as:
```suggestion
        kv := map[string]string{"dstHost":"www.example.com"}
	if _, ok := kv["dstHost"]; !ok {
		fmt.Print("fail")
	}
	fmt.Println(kv["dstHost"])
```

rocwenlinux: The print "fmt.Println("hosts", md)" is only for debug purpose, it shouldn't be enabled/included in production. Otherwise each request will trigger this print, cause the performance issue.

codyja:
We hit the 4mb limit also. Any chance we can get this merged in and a release cut? Thanks!

nathanejohnson:
@rocwenlinux can you squash this?  If so  I'll merge this and slate it for 1.6

rocwenlinux:
> @rocwenlinux can you squash this? If so I'll merge this and slate it for 1.6

squashed.

main.go
```diff
@@ -161,6 +161,8 @@ func newGrpcProxy(cfg *config.Config, tlscfg *tls.Config) []grpc.ServerOption {
 		grpc.UnknownServiceHandler(handler),
 		grpc.StreamInterceptor(proxyInterceptor.Stream),
 		grpc.StatsHandler(statsHandler),
```
nathanejohnson, line 163: It would be great to have these values be configurable.  Could you add these to config/config.go in the Config struct, maybe as GRPCMaxRxMsgSize and GRPCMaxTxSize , and then setting sane default values in config/default.go ?  Perhaps setting the current default of 4MB is still sane here, 1gb seems like it might be excessive for many people?  And if it's configurable, this allows people to relax the limits.

proxy/grpc_handler.go
```diff
@@ -231,7 +251,7 @@ func (p *grpcConnectionPool) Get(ctx context.Context, target *route.Target) (*gr
 
 func (p *grpcConnectionPool) newConnection(ctx context.Context, target *route.Target) (*grpc.ClientConn, error) {
 	opts := []grpc.DialOption{
-		grpc.WithDefaultCallOptions(grpc.CallCustomCodec(grpc_proxy.Codec())),
```
nathanejohnson, line 234: See above about the configurable limits

nathanejohnson (changes requested):
Can we make max message sizes configurable?

nathanejohnson:
I'm going to merge this and do a subsequent PR to make these limits configurable.  Some good work in here and I'm guessing the original authors have moved on.

PR #637, opened 2019-04-12
fix exit status code

ianic:
Fabio always returned status code 1. Even on clean exit.
Shutdown of http server is treated as fatal:
 ```
[FATAL] http: Server closed
[FATAL] ui: http: Server closed
```

HTTP server [Serve](https://golang.org/pkg/net/http/#Server.Serve) method always returns non nil error.

ErrServerClosed should be treated as clean exit.
> After Shutdown or Close, the returned error is ErrServerClosed."


pschultz:
Looks good to me.

leprechau:
agreed, good catch

PR #639, opened 2019-04-12
use vendor directory when building - fixes #638

### LINKED ISSUE #638: Build should use included vendor directory with modules
The current build process bypasses the vendor directory when building with go module support.  Since we include the vendor directory we should tell go to use it when building.

leprechau:


leprechau:
Anyone see any problems with this?  It does force using golang 1.11+ to build but I think we're already there.

pschultz:
Go 1.10 and earlier are not supported by the maintainers since 1.12 has been released, so dropping support here is totally fine, I think.

Since Go <= 1.8 won't work anymore, [you can drop the various `| grep -v '/vendor/'` bits](https://golang.org/doc/go1.9#vendor-dotdotdot) while you're at it.

leprechau:
Yeah I caught that ... just didn't get it in that last commit

PR #641, opened 2019-04-15
ui: Remove duplicate destination column

### LINKED ISSUE #628: Route table UI is broken
Hello,

it looks like v1.5.11 has a broken UI which displays route table: it has an additional column with url, but missing header (so 6 columns in header and 7 in body).

See:
https://github.com/fabiolb/fabio/blob/master/admin/ui/route.go#L84
https://github.com/fabiolb/fabio/blob/master/admin/ui/route.go#L99

pschultz:
#587 duplicated the destination column in the route UI instead of replacing it, probably as a result of resolving the merge conflict with #588.

Fixes #628.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,11 +80,11 @@
 
   Thanks to [@stack72](https://github.com/stack72) for the patch.
 
-* [PR #583](https://github.com/fabiolb/fabio/pull/583): Make dest column clickable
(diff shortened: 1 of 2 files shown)


scalp42:
@pschultz any chance to see this merged?

scalp42:
Thanks a lot @pschultz :v

PR #645, opened 2019-04-18
ensure absolute path after strip to maintain rfc complaince

leprechau:
It's easy to inadvertently remove the leading slash from the request path when stripping which results in sending a malformed GET with a relative path to the backend server.

Example below ...

Origin URL:
```
http://foo.com/apples/oranges/steve.html
```

Route:
``` 
route add strippy /apples/ http://1.2.3.4:80/ opts "strip=/apples/"
```

Results in sending `GET oranges/steve.html` to the 1.2.3.4 backend.  Proper GET request should always be an absolute path or absolute URI according to the RFC (https://tools.ietf.org/html/rfc2616#section-5.1.2).


leprechau:
@nathanejohnson This should resolve your issue.

leprechau:
Got it, I'll update the comment.  Thanks.

Issue #647, opened 2019-04-19
NormalizeHost CPU Performance


murphymj25:
In looking at a CPU pprof of one of our production instances, we have noticed that a significant amount of CPU time is being spend on the strings.ToLower in the normalizeHost function in table.go.

The pprof was taken on a 16 core instance, with about 3,000 requests per second of traffic.  In this environment, the route table has about 13,000 separate routes.  In the pprof, we see the following:
![image](https://user-images.githubusercontent.com/36900417/56437253-48613700-62a4-11e9-8a7e-f1e83674ee1e.png)

I've done some testing and updated the matchingHostNoGlob to use string.EqualFold instead of doing the comparison with the normalizeHost and strings.ToLower.  With that change, the normalizeHost was reduced from 22% to 8% with similar request volume and route table size.

I'll create a PR to share the changes I've done already.  I think there's still some opportunity to further reduce the CPU load by changing how the :port is trimmed off.

PR #648, opened 2019-04-19
Issue 647 NormalizeHost

murphymj25:
#647

### DIFF
--- a/route/table.go
+++ b/route/table.go
@@ -314,6 +314,16 @@ func normalizeHost(host string, tls bool) string {
 	return host
 }
 
+func normalizeHostNoLower(host string, tls bool) string {
+	if !tls && strings.HasSuffix(host, ":80") {
+		return host[:len(host)-len(":80")]
+	}
+	if tls && strings.HasSuffix(host, ":443") {
+		return host[:len(host)-len(":443")]
+	}
+	return host
+}
+
 // matchingHosts returns all keys (host name patterns) from the
 // routing table which match the normalized request hostname.
 func (t Table) matchingHosts(req *http.Request) (hosts []string) {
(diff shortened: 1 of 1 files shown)


murphymj25:
For this PR, i've created a new NormalizeHostNoLower which doesn't have the string.ToLower.  And changed the if statement in matchingHostNoGlob to use string.EqualFold.  This allows the comparison to happen without needing to lower the case.

I don't think we would need to have the separate function, but this approach allowed me to quickly test the change for the matchigHostNoGlob function, which is what we are using.

As I stated in #647 I think there is some opportunity to further reduce the CPU utilization by changes port suffix are trimmed off or perhaps accounting for this right in the matching host functions.  

Any thoughts on how best to approach this would be appreciated.

pschultz:
https://github.com/fabiolb/fabio/blob/660ad7605415e74a0d73954521bd3eb3d0d4a35d/route/table.go#L158-L160
https://github.com/fabiolb/fabio/blob/660ad7605415e74a0d73954521bd3eb3d0d4a35d/route/table.go#L184

The table keys are already lowercase, so it seems we don't even need to use EqualFold:

```go
func (t Table) matchingHostNoGlob(req *http.Request) (hosts []string) {
    host := normalizeHost(req.Host, req.TLS != nil) // note: *with* ToLower here

    for pattern := range t { // pattern is already lower case
        normpat := normalizeHostNoLower(pattern, req.TLS != nil)
        if normpat == host {
            hosts = append(hosts, pattern)
            return
        }   
    }   
    return
}
```

The Route.Glob field is used to cache compiled path globs. It seems simple enough to add a field for the normalized host, so we don't have to call normalizeHostNoLower on lots of never-changing values for each request. If you want to improve performance with globs enabled, you could also add another field for pre-compiled host globs.

route/table.go
```diff
@@ -314,6 +314,16 @@ func normalizeHost(host string, tls bool) string {
 	return host
 }
 
+func normalizeHostNoLower(host string, tls bool) string {
```
pschultz: You should change `normalizeHost` to use this new function:

```go
func normalizeHost(host string, tls bool) string {
    return strings.ToLower(normalizeHostNoLower(host, tls))
}
```

murphymj25:
@pschultz Thanks for the feedback.  It makes sense to remove the equalfold and just do a tolower on the host.  I'll get this updated.

leprechau:
@pschultz @murphymj25 We good to merge this one?  It looks like the comments have been resolved.

murphymj25:
From my perspective this should be ready to merge.  We have been running it in our environments since May without any issues encountered.

leprechau:
@murphymj25 That works for me.  Thank you again.

PR #470, opened 2018-03-18
Issue #460: Fix access logging when gzip is enabled

magiconair:
The code that captured the content length and the status code
for the access logger depended on the http.Handler to be of a
certain type. Wrapping the handler in a gzip handler broke this
silently.

This fix removes that dependency and instead wraps the response
writer directly to capture the values required for access logging.

Fixes #460 
Closes #469

PR #655, opened 2019-05-06
Fix SSE by implementing Flusher in responseWriter wrapper

gfloyd:
I noticed after upgrading to any version later than 1.5.8 that server-sent events stopped working. It looks like the responseWriter wrapper added in #470 does not implement the `http.Flusher` interface, so events are never flushed to the client.

Adding `Flush()` to the wrapper fixes the issue and server-sent events are working for me again.


leprechau:
Thank you, nice catch!

PR #658, opened 2019-05-16
config: let registry.consul.register.addr default to ui.addr

### LINKED ISSUE #657: Wrong reported HealthCheck-URI using custom -proxy.addr & -ui.addr
When starting fabio with custom port numbers, i.e.:  fabio1.5.exe -proxy.addr :8888 -ui.addr :8887,  the health check URI reported to Consul is wrong.

Current output:
< Registered fabio with health check to "http://[my-private-ip]:9998/health >

BugFix:  the proxy.add parameter should be considered in the URL passed to Consul.

pschultz:
Always set registry.consul.register.addr to the value of ui.addr if the
flag hasn't been specified.

Fixes #657.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,22 @@
 
 ### Unreleased
 
+#### Breaking Changes
+
+* [Issue #657](https://github.com/fabiolb/fabio/issues/657): Fix default registration address
+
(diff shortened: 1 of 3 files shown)


leprechau:
LGTM!

PR #674, opened 2019-07-15
Deprecate deregisterCriticalServiceAfter option

pschultz:
The deregisterCriticalServiceAfter option applied to the HTTP health
check for services that are registered in Consul.

#664 introduced additional Consul TTL checks. These TTL
checks have DeregisterCriticalServiceAfter set to the minimum of one
minute and will therefore always beat or tie the HTTP check's
DeregisterCriticalServiceAfter option, rendering it obsolete.

Document this fact in the usage message and HTML docs.


registry/consul/register.go
```diff
@@ -167,5 +170,5 @@ func serviceRegistration(cfg *config.Consul, serviceName string) (*api.AgentServ
 }
 
 func computeServiceTTLCheckId(serviceID string) string {
-	return strings.Join([]string{serviceID, "ttl"}, "-")
+	return serviceID + "-ttl"
 }
```
leprechau: I'd like to keep the `strings.Join` function and eventually convert other plain string concatenations to the strings functions to avoid potential leaked allocations and reduce unnecessary garbage collection.

Quick test case of the small performance hit vs allocation save can be seen here: https://gist.github.com/dtjm/c6ebc86abe7515c988ec

pschultz: I don't follow. I changed it precisely for performance (and readability). The slice must be allocated but there is no allocation necessary for the string concatenation:

```go
// registry/consul/register_test.go
package consul

import (
        "strings"
        "testing"
)

func checkIDJoin(serviceID string) string {
        return strings.Join([]string{serviceID, "ttl"}, "-")
}

func checkIDConcat(serviceID string) string {
        return serviceID + "-ttl"
}

func BenchmarkJoin(b *testing.B) {
        for i := 0; i < b.N; i++ {
                s := checkIDJoin("my-service")
                _ = s
        }
}

func BenchmarkConcat(b *testing.B) {
        for i := 0; i < b.N; i++ {
                s := checkIDConcat("my-service")
                _ = s
        }
}
```

```txt
$ go version
go version go1.12.5 linux/amd64
$ go test -bench=. -benchmem -run=XXX ./registry/consul
goos: linux
goarch: amd64
pkg: github.com/fabiolb/fabio/registry/consul
BenchmarkJoin-8         20000000                60.9 ns/op            16 B/op          1 allocs/op
BenchmarkConcat-8       50000000                25.1 ns/op             0 B/op          0 allocs/op
PASS
ok      github.com/fabiolb/fabio/registry/consul        2.570s
```

I'm not sure why some people in that gist report concatenation being equal to or worse than strings.Join. I can't reproduce that. I've tried much longer strings, randomly generated strings (to rule out any kind of optimization related to constants), and multiple concatenations. Concat always beats strings.Join for me.

I don't care enough to fight over this though.

leprechau: Sorry for leaving this hanging.

leprechau:
Re-running tests via Travis integration.

PR #675, opened 2019-07-16
Fix matching priority for host:port tuples

### LINKED ISSUE #656: Clarify documentation HTTP Redirects
Tried to redirect http to https with service tags. I followed the documentation https://fabiolb.net/feature/http-redirects/ which says:

> To redirect from HTTP to HTTPS you must include the host:port of the HTTP endpoint

It did not work. I also have to include the port for the HTTPS rule.

Did not work:

> urlprefix-web1.traefik/,urlprefix-web1.traefik:80/ redirect=301,https://web1.traefik/

Did work:

> urlprefix-web1.traefik:443/,urlprefix-web1.traefik:80/ redirect=301,https://web1.traefik/

It surely is clear for users which uses fabiolb for some time, but I just evalute fabiolb, so for me was it not clear. I found some issues about this topic, so perhaps the documentation can be a bit clearer about redirecting.

pschultz:
e566aaab reversed DNS names since they have
their most significant part at the front. Unfortunately it didn't
consider cases where the name is followed by a :port, making names
without port always take precedence over names with a port.

This patch changes the reverse function to ignores such port suffixes.

Closes #656

### DIFF
--- a/route/table.go
+++ b/route/table.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"net"
 	"net/http"
 	"net/url"
 	"sort"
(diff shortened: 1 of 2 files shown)


leprechau:
Good catch.  LGTM.  Thank you!

PR #677, opened 2019-07-24
Add Vault example to the traffic shaping section.

jrasell:
Add a small section to the Traffic Shaping documentation detailing how to use Fabio to route traffic for Vault to the active node.

Ref: https://twitter.com/jrasell/status/1100724689767485440


pschultz:
Thank you!

PR #692, opened 2019-09-04
Implement new config to override service host with consul advertise ip

harlequin:
Nomad in combination with Consul often register a service with an internal IP
Nomad doesn't respect to register services with the advertisement IP

With this setting all services host IP will be ignored and the consul advertise IP will be registered.


leprechau:
@harlequin Why is the service address empty?  This shouldn't be empty if Nomad is doing the service registration.

https://www.nomadproject.io/docs/job-specification/service.html

This is how we're using Nomad with Consul and Fabio today.

PR #710, opened 2019-10-11
Safely handle missing cert from Vault KV store

dradtke:
I have certificates saved in Vault using its KV engine version 2, and Fabio started panicking after I deleted an unused certificate. The fix was to run `vault kv metadata delete <cert>` to clean up the metadata as well, but Fabio should be able to avoid panicking in this case.


leprechau:
Good catch, thank you!

PR #724, opened 2019-11-24
add path

spiderDetective:
fabio supports stripping a path from the incoming request. If you want to forward http://host/foo/bar as http://host/bar you can add a strip=/foo option to the route options as urlprefix-/foo/bar strip=/foo. But, sometimes, need to add pre path.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -108,3 +108,5 @@ require (
 )
 
 go 1.13
+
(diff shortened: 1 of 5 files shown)


leprechau:
@spiderDetective could you please sign the CLA and add documentation for this new route option?

PR #739, opened 2019-12-27
Fix infinite buffering of SSE responses when gzip is enabled

ctlajoie:
SSE requests (`text/event-stream`) don't work properly (no events are transmitted to the client) if `proxy.gzip.contenttype` is non-nil and gzip is in the `Accept-Encoding` header, because under those conditions the response gets chained through `GzipResponseWriter` even if no compression is being done. This change will bypass the `GzipResponseWriter` if there is a `Accept: text/event-stream` header.

### DIFF
--- a/proxy/gzip/gzip_handler.go
+++ b/proxy/gzip/gzip_handler.go
@@ -21,12 +21,15 @@ import (
 
 const (
 	headerVary            = "Vary"
+	headerAccept          = "Accept"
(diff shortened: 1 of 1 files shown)


leprechau:
Triggering a rebuild now that Travis is functional again.

leprechau:
@ctlajoie Thank you! Sorry again for the delay.

PR #741, opened 2020-01-22
Use proxy protocol library that supports v2

SarahRiley:


marco-m:
Hello @Sarah-E-Greene, do you have any news about this upgrade to Proxy Protocol v2 ? (See also: #1000)
Thank you.

PR #757, opened 2020-03-20
Add `only=` tag

rigrassm:
The new tag `only=` allows the user to have more granular control over which checked services
get added to the routing table.

Example:
Nomad publishes service checks for the HTTP, RPC, and Serf services to consul. Registering the tag below results in
a routing table that splits requests for nomad.service.example.com between the three different protocols.

Service Tag
```
    "urlprefix-nomad.service.example.com:443/ proto=https tlsskipverify=true"
```

Resulting Routing Table Entries
```
nomad        nomad.service.example.com:443/        https://172.16.1.8:4646        tlsskipverify=true        33%
nomad        nomad.service.example.com:443/        https://172.16.1.8:4647        tlsskipverify=true        33%
nomad        nomad.service.example.com:443/        https://172.16.1.8:4648        tlsskipverify=true        33%
```

Since we really only want to route the HTTP traffic, this commit allows you to use the rule below to only generate
a single routing table entry based on the port number associated with our HTTP port.

Service Tag
```
    "urlprefix-nomad-0.service.example.com:443/ proto=https only=4646 tlsskipverify=true"
```

Resulting Routing Table Entries
```
nomad        nomad.service.example.com:443/        https://172.16.1.8:4646        tlsskipverify=true        100.00%
```

### DIFF
--- a/registry/consul/routecmd.go
+++ b/registry/consul/routecmd.go
@@ -55,6 +55,7 @@ func (r routecmd) build() []string {
 
 			var weight string
 			var ropts []string
+			var only string
 			for _, o := range strings.Fields(opts) {
 				switch {
 				case o == "proto=tcp":
@@ -68,6 +69,9 @@ func (r routecmd) build() []string {
 
 				case o == "proto=grpc":
 					dst = "grpc://" + addr
+
+				case strings.HasPrefix(o, "only="):
+					only = o[len("only="):]
 
 				case strings.HasPrefix(o, "weight="):
 					weight = o[len("weight="):]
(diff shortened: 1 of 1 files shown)


rigrassm:
@nathanejohnson Sorry, I had signed the CLA but had accidentally messed up the email address in one of the commits so it wasn't registering. Went ahead and squashed everything down to 1 commit and re-signed the CLA so it should be good now.

nathanejohnson:
@rigrassm tentatively this looks good, could you add a test case in routecmd_test.go to cover this option.  Now the interesting part here is this means consul registry will have a route option that doesn't exist otherwise.  On the other hand, filtering doesn't make much sense in other cases.  I went looking for documentation of the consul tags and it's not really there.  I would have asked you to update that documentation, but since it doesn't exist I'll make it a point to create that page and roll it out with the next release.

TLDR; add a test case and I'll merge this.

tristanmorgan:
for future reference one way to achieve routing to only the http port for Nomad is to add a weighting rule like the following to the Consul KV under fabio/config...
```
route weight nomad nomad.service.example.com weight 1.00 tags "http"
```

Issue #691, opened 2019-09-04
Simple HTTP path prefix replacement


JamesJJ:
I'd like to be able to do simple HTTP path substitution e.g. path prefix replacement, however I'm not sure which way would be most appropriate.

e.g `GET /aa/bb/cc ---> GET /zz/bb/cc`

When HTTP forwarding, the `opts "strip=/some_path"` option facilitates removal of some existing path prefix. Adding a replacement prefix could be implemented with a new `opts add=/new_path`, or taken directly from the route `dst`.

Here is a P.O.C. that _works for me_ based upon directly using the path specified in `dst`:
https://github.com/JamesJJ/fabio/commit/09b7e3f1782554eaea141842617c2a68bbc0b46f

I'd appreciate any comments or advice. If this direction is acceptable, I'll see about tests+docs and send a PR.

PR #767, opened 2020-05-08
Simple HTTP path prefix replacement

JamesJJ:
This is a potential solution for #691 

A new option `prepend` is introduced. For example, `prepend=/baz` allows an incoming request with path `/bar` to be forwarded upstream with path `/baz/bar`.

Combining `prepend=/baz` with the existing `strip=/foo` option, allows an incoming request path `/foo/bar` to be forwarded upstream with path `/baz/bar`.

This change should not affect any existing functionality, and the docs are updated to match.

### DIFF
--- a/docs/content/cfg/_index.md
+++ b/docs/content/cfg/_index.md
@@ -29,6 +29,7 @@ Option                                     | Description
 `allow=ip:10.0.0.0/8,ip:fe80::/10`         | Restrict access to source addresses within the `10.0.0.0/8` or `fe80::/10` CIDR mask.  All other requests will be denied.
 `deny=ip:10.0.0.0/8,ip:fe80::1234`         | Deny requests that source from the `10.0.0.0/8` CIDR mask or `fe80::1234`.  All other requests will be allowed.
 `strip=/path`                              | Forward `/path/to/file` as `/to/file`
+`prepend=/prefix`                          | Forward `/path/to/file` as `/prefix/path/to/file`
 `proto=tcp`                                | Upstream service is TCP, `dst` must be `:port`
 `pxyproto=true`                            | Enables PROXY protocol on outbount TCP connection
 `proto=https`                              | Upstream service is HTTPS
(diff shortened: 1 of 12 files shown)


lizongshen:
any progress？

lizongshen:
This function is very useful. When can I expect to use it？

lizongshen:
I added prepend=baz tag to consult services，but it doesn't work?

ketzacoatl:
@nathanejohnson, can you elaborate on the `thinking` label? What is needed to make a decision on this?

bac-w:
@nathanejohnson the functionality is very useful, is it possible to see it in the next release?

leprechau:
There are a few conflicts in this branch due to the fix for #824 that was recently pushed. The `prepend` should probably also happen before replacement of the pseudo variable.

JamesJJ:
@leprechau Thank you for checking on this one. I've resolved the conflicts.

I'm not totally clear about 「The prepend should probably also happen before replacement of the pseudo variable」 -  I'd be grateful if you could explain further if this needs other adjustments.

leprechau:
@JamesJJ Could you take a look at #824 and see if the changes there do the same thing as the prepend option?  I think these changes allow the same behavior.

JamesJJ:
@leprechau It's been a while since I played with this so I may be wrong, however my interest was in the http proxy function (not http redirect). It looks to me that between https://github.com/fabiolb/fabio/blob/master/proxy/http_proxy.go#L133 where the target path is set to the incoming request path, and https://github.com/fabiolb/fabio/blob/master/proxy/http_proxy.go#L198 where the request is proxied out, only `StripPath` updates the target path.

So if I read it correctly, `t.URL.Path` is not utilised when proxying - so we need a new `prepend` option, or update to involve `t.URL.Path`.

I think I opted for a new option as there is no risk of a breaking change if anyone is currently relying on the target URL path being ignored when proxying.

leprechau:
@JamesJJ Thank you, that makes sense. You are correct the changes I was mentioning only affect redirects whereas this options is implemented in both the redirect and straight proxy portions of the code.

Thank you again, LGTM!

Issue #783, opened 2020-08-28
Add support for HTTPS+TCP+SNI on the same listener


nathanejohnson:
In some cases it is desirable to have TCP SNI steering as well as HTTPS proxying.

For example: Say you have one external IP address, 1.2.3.4

You have a services A, B and C that need path based (https) routing

You have service D that needs to see the client certificate, and any traffic with this SNI should be routed to D so it can terminate TLS

Adding a https+tcp+sni mode would support this use case by allowing any hosts that match D - as specified by proto=tcp or scheme is tcp:// in the route - would be tcp proxied, and that everything else would fall through to existing HTTPS routing logic.

PR #784, opened 2020-08-28
https+tcp+sni listener support

nathanejohnson:
Adds support for the https+tcp+sni listener.  Any host routes that are marked with
proto=tcp or have a scheme tcp:// will be matched for TCP SNI steering.  Failing
any matches there, fallthrough will be to https matching.  This resolves: #783.

clean up error handling on 'use of closed network connection' case

fix makefile tag finding stuff

fix logic for https+tcp+sni matching so that only explicit proto=tcp matches


nathanejohnson:
I'll give this a few days, failing any negative reviews I'm merging this

proxy/inetaf_tcpproxy.go
```diff
@@ -0,0 +1,104 @@
...
+type InetAfTCPProxyServer struct {
+	Proxy    *tcpproxy.Proxy
+	children []*childProxy
+}
+
+// Close - implements Server - is this even called?
+func (tps *InetAfTCPProxyServer) Close() error {
+	_ = tps.Proxy.Close()
+	firstErr := tps.Proxy.Wait()
+	errChan := make(chan error)
```
dcarbone, line 28: Shouldn't this be buffered, as it is in the `.Serve()` method below?  also probably a good idea to add a `defer close(errChan)`

proxy/inetaf_tcpproxy.go
```diff
@@ -0,0 +1,104 @@
...
+
+}
+
+// Serve - implements server.  The listener is ignored, but it
+// calls serve on the children
+func (tps *InetAfTCPProxyServer) Serve(_ net.Listener) error {
+	if len(tps.children) == 0 {
+		return fmt.Errorf("no children defined for listener")
+	}
+	errChan := make(chan error, len(tps.children))
```
dcarbone, line 56: would be good to add a defer close here, too.

proxy/inetaf_tcpproxy.go
```diff
@@ -0,0 +1,104 @@
...
+// tcpproxy.TargetListener or a derivative.  Don't call after
+// Serve() is called.
+func (tps *InetAfTCPProxyServer) ServeLater(l net.Listener, s Server) {
+	tps.children = append(tps.children, &childProxy{l, s})
+}
+
+func (tps *InetAfTCPProxyServer) Shutdown(ctx context.Context) error {
+	_ = tps.Proxy.Close()        // always returns nil error anyway
+	firstErr := tps.Proxy.Wait() // wait for outer listener to close before telling the childProxy
+	errChan := make(chan error, len(tps.children))
```
dcarbone, line 88: defer close here, as well.

proxy/inetaf_tcpproxy_integration_test.go
```diff
@@ -0,0 +1,177 @@
...
+		)
+		if proto, ok = t.Opts["proto"]; !ok && t.URL != nil {
+			proto = t.URL.Scheme
+		}
+		return "tcp" == proto
+	}
+
+	// get an unused port for use for the proxy.  the rest of the tests just
+	// pick a high-numbered port, but this should be safer, if ugly.  could
+	// also just fire up a listener with 0 as the port and let the stack
```
dcarbone, line 82: this is what hashi used to do for consul testing, as well, however it doesn't work on all os's (this is actually one of the first things @magiconair fixed properly when he consulted for them)

old: https://github.com/hashicorp/consul/blob/v0.8.5/testutil/server.go#L132

current: https://github.com/hashicorp/consul/tree/v1.8.3/sdk/freeport

while i don't think its worth messing with now, may be nice to keep in the back of our minds if a user runs into problems while executing tests.

proxy/inetaf_tcpproxy.go
nathanejohnson: the defer close would not be necessary, but I agree it should be consistent with the other buffered error channels elsewhere.

proxy/inetaf_tcpproxy.go
nathanejohnson: see above

.travis.yml
```diff
@@ -6,6 +6,11 @@ go:
   - 1.13
```
dcarbone, line 6: Lets add `1.14` and `1.15` here.

proxy/inetaf_tcpproxy.go
nathanejohnson: see above

proxy/inetaf_tcpproxy_integration_test.go
nathanejohnson: your after link doesn't work, but I'd be interested to see what they come up with.  This test skips unless you explicitly create the /etc/hosts entries and set the environment variable mentioned in the comments above, but as far as I'm aware this should work cross platform provided it resolves example.com and example2.com to localhost.  So are you saying httptest is broken on some platforms?

dcarbone: @nathanejohnson: it works for me...?

nathanejohnson: I see the package, I assumed you were trying to point to a specific section of code as in the first link?  Also, is httptest broken on some platforms?  Because that's what I use to grab a free port

dcarbone: No, I linked the package as it contains os-specific implementations. And no, i haven't encountered it personally.

Issue #786, opened 2020-09-09
matchingHostNoGlob sometimes returns incorrect matched host


nathanejohnson:
related to https://github.com/fabiolb/fabio/issues/506 , it seems that matchingHostNoGlob doesn't do this reverse sort.  It just picks the first one it finds.

PR #787, opened 2020-09-09
Fixing issue #786 - matchingHostNoGlob sometimes returns incorrect host

nathanejohnson:
Moves logic for reversing hosts / ports to get most specific match into
separate method

Calls this same method from both matchingHost and matchingHostNoGlob


leprechau:
LGTM

PR #789, opened 2020-09-15
update version to a dev version

shantanugadgil:
Now that v 1.5.14 is GA, would it make sense to mark the version as a "dev" version.

_I tried to go through the Makefile, but didn't find the usual ldflag value trick usually used
to send the version string from outside_

@nathanejohnson

### DIFF
--- a/main.go
+++ b/main.go
@@ -50,7 +50,7 @@ import (
 // It is also set by the linker when fabio
 // is built via the Makefile or the build/docker.sh
 // script to ensure the correct version number
-var version = "1.5.14"
+var version = "1.5.15-dev"
 
 var shuttingDown int32
 


nathanejohnson:
@shantanugadgil could you clarify what you mean?

shantanugadgil:
I mean that any intermediate builds until 1.5.15 is GA will show the correct "dev" string as part of the version.

nathanejohnson:
@shantanugadgil  This is a good idea, I'm just making sure I have a handle on how this Makefile works fully before I merge.

nathanejohnson:
@shantanugadgil so the make release step expects to make a tag, and since a -dev version would be a moving target, maybe it would make more sense to have the git branch and a short sha of the head commit?  Unfortunately just updating that one file does not mean make build will honor your "version".  But as a temporary workaround, you can set an arbitrary version number like so:

`CUR_TAG=v1.5.15-dev make build`

nathanejohnson:
@shantanugadgil so this is what is used in the current system:

`git describe --tags --first-parent`

Right now, at master, it gives this:

`v1.5.14-7-g231e7f2`

which means 7 commits since the v1.5.14 tag, and abbreviated head sha is 231e7f2.  The g prefix is described in the docs here:

```
The "g" prefix stands for "git" and is used to allow describing the version of a
       software depending on the SCM the software is managed with. This is useful in an environment where people may use different SCMs.
```

shantanugadgil:
hmm ... ok, if this is already handled, I can close this.

PR #798, opened 2020-10-30
Stop dynamic TCP listener when upstream is no longer available

### LINKED ISSUE #796: TCP Dynamic Proxy is not releasing ports from deregistered services
### Abstract
`fabio` leaves dangling ports from deregistered services that used _TCP Dynamic Proxy_

### Steps to reproduce
1. Start `fabio:1.5.14-go1.15` docker container with `FABIO_PROXY_ADDR=":9999,0.0.0.0:0;proto=tcp-dynamic;refresh=5s"` and [host networking](https://docs.docker.com/network/host/).
2. Register 2 consul services that will use _TCP Dynamic Proxy_
```
netstat -tulpn | grep fabio                                                                                                                                                          nomad-1: Tue Oct 27 16:37:20 2020

tcp6	   0	  0 :::31130                :::*                    LISTEN	29060/fabio
tcp6	   0	  0 :::24604                :::*                    LISTEN	29060/fabio
tcp6	   0	  0 :::9998                 :::*                    LISTEN	29060/fabio
tcp6	   0	  0 :::9999                 :::*                    LISTEN	29060/fabio
```
3. Deregister one of the consul services. Port `31130` is dangling, despite the fact that the underlying service is deregistered and no longer visible on Fabio UI (http://foobar.fabio.ui:9998/routes)
```
netstat -tulpn | grep fabio                                                                                                                                                          nomad-1: Tue Oct 27 16:37:20 2020

tcp6	   0	  0 :::31130                :::*                    LISTEN	29060/fabio
tcp6	   0	  0 :::24604                :::*                    LISTEN	29060/fabio
tcp6	   0	  0 :::9998                 :::*                    LISTEN	29060/fabio
tcp6	   0	  0 :::9999                 :::*                    LISTEN	29060/fabio
```
4. Restart `fabio` container to see if it will release ports and start properly. 
```
netstat -tulpn | grep fabio                                                                                                                                                          nomad-1: Tue Oct 27 16:37:20 2020

tcp6	   0	  0 :::24604                :::*                    LISTEN	29060/fabio
tcp6	   0	  0 :::9998                 :::*                    LISTEN	29060/fabio
tcp6	   0	  0 :::9999                 :::*                    LISTEN	29060/fabio
```

fwkz:
This PR closes #796

### DIFF
--- a/proxy/serve.go
+++ b/proxy/serve.go
@@ -28,23 +28,21 @@ type Server interface {
 var (
 	// mu guards servers which contains the list
 	// of running proxy servers.
-	mu                sync.Mutex
-	servers           []Server
-	dynamicProxyPorts = make(map[string]Server)
+	mu      sync.Mutex
+	servers = make(map[string]Server)
 )
 
-func CloseDynamicProxy(port string) error {
-	if srv, ok := dynamicProxyPorts[port]; ok {
+func CloseProxy(address string) error {
+	mu.Lock()
+	if srv, ok := servers[address]; ok {
 		err := srv.Close()
 		if err != nil {
 			return err
 		}
-		log.Printf("[INFO] Dynamic TCP listener on port %s has been terminated", port)
-		mu.Lock()
-		servers = removeServer(servers, srv)
-		delete(dynamicProxyPorts, port)
-		mu.Unlock()
+		log.Printf("[INFO] Dynamic TCP listener on %s has been terminated", address)
+		delete(servers, address)
 	}
+	mu.Unlock()
 	return nil
 }
 
@@ -53,15 +51,17 @@ func Close() {
 	for _, srv := range servers {
 		srv.Close()
 	}
-	servers = []Server{}
+	servers = make(map[string]Server)
 	mu.Unlock()
 }
 
 func Shutdown(timeout time.Duration) {
 	mu.Lock()
-	srvs := make([]Server, len(servers))
-	copy(srvs, servers)
-	servers = []Server{}
+	srvs := make(map[string]Server, len(servers))
+	for k, v := range servers {
+		srvs[k] = v
+	}
+	servers = make(map[string]Server)
 	mu.Unlock()
 
 	var wg sync.WaitGroup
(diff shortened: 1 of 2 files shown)


proxy/serve.go
```diff
@@ -177,3 +197,14 @@ func serve(ln net.Listener, srv Server) error {
 	}
 	return err
 }
+
+func removeServer(srvs []Server, s Server) []Server {
+	for i, srv := range srvs {
+		if s == srv {
```
nathanejohnson, line 203: I'm a bit concerned this is going to be brittle.  A better, if more invasive solution might be to tweak the interface for Server here so that we can add a method to return the listen address, and then compare the returns from the Address() methods.  We need to add this method to the existing implementations of course.

nathanejohnson (changes requested):

proxy/serve.go
```diff
@@ -157,6 +174,9 @@ func ListenAndServeTCP(l config.Listen, h tcp.Handler, cfg *tls.Config) error {
 		ReadTimeout:  l.ReadTimeout,
 		WriteTimeout: l.WriteTimeout,
 	}
+	if _, ok := h.(*tcp.DynamicProxy); ok {
```
nathanejohnson, line 177: this should be protected by a mutex here

proxy/serve.go
nathanejohnson: on second thought, I'm thinking that instead of changing the interface to have an Address() method, maybe just extend the use of your map to include all servers, i.e., change servers []Server to servers map[string]Server , and the key is the listen address .  Thoughts?

fwkz: If we change `servers` from slice to map what about `ListenAndServeHTTPSTCPSNI()` which passes `nil` as a listener? Should we pass the result of `tls.NewListener(httpsListener, cfg)` (line no. 139) anyway just to extract address and use it as a key in the `servers` map?

nathanejohnson: That's certainly an option.  In that section, httpsListener, the TLS wrapped version of httpsListener that's passed into ServeLater() and tps are net.Listeners which will return the same Addr() .  Another option is to avoid using the Addr() method of net.Listener and just add a string parameter to the serve() method.

proxy/serve.go
```diff
@@ -28,23 +29,39 @@ var (
 	// mu guards servers which contains the list
 	// of running proxy servers.
 	mu      sync.Mutex
-	servers []Server
+	servers = make(map[string]Server)
 )
 
+func CloseProxy(address string) error {
+	if srv, ok := servers[address]; ok {
```
fwkz, line 36: Shouldn't we lock the whole `if` statement since we are doing read+write? 🤔

nathanejohnson: Yes

fwkz: Done

nathanejohnson:
Tentatively this looks good, going to pull this down and test it locally today.  If all goes well I'll merge this ASAP.

nathanejohnson (approved):

nathanejohnson:
@fwkz thank you for your contribution!

fwkz:
> @fwkz thank you for your contribution!

I'm glad that I could help. 💪

PR #806, opened 2020-12-02
Updating tcp dynamic proxy to match on routes that are port only

### LINKED ISSUE #797: TCP Dynamic Proxy route without specifying exact IP?
I'm using tcp-dynamic proxy with `proxy.addr = :9999,0.0.0.0:0;proto=tcp-dynamic;refresh=5s` but wondering if there's a way to use routes without hardcoding an IP address if I understand correctly?

For example, I have a service with the following tags:

```
fabio-0.0.0.0:6379 proto=tcp
fabio-:6379 proto=tcp
fabio-mydnsrecord:6379 proto=tcp
```

The routes get added successfully, but they all result in "No route found" logs.

```
Oct 30 18:07:52 default-ubuntu-1804 fabio[19887]: 2020/10/30 18:07:52 [WARN] No route for 192.168.33.33:6379
```

Is it possible to be able to resolve a hostname instead of use an IP address, or route based on port-only instead of ip:port?

Thanks in advance!

nathanejohnson:
This fixes #797

### DIFF
--- a/proxy/tcp/tcp_dynamic_proxy.go
+++ b/proxy/tcp/tcp_dynamic_proxy.go
@@ -36,8 +36,13 @@ func (p *DynamicProxy) ServeTCP(in net.Conn) error {
 	if p.Conn != nil {
 		p.Conn.Inc(1)
 	}
+
 	target := in.LocalAddr().String()
 	t := p.Lookup(target)
+	if t == nil {
+		_, port, _ := net.SplitHostPort(target)
+		t = p.Lookup(":" + port)
+	}
 	if t == nil {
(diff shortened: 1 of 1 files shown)


nathanejohnson:
@murphymj25 could you take a look at this?

Thanks!

murphymj25:
@nathanejohnson It's been a while since I've looked at this code, but based tcp_proxy.go what you have looks correct.

killerdejavu:
@nathanejohnson Would be great to have this PR merged! 😄

leprechau:
LGTM, and thanks @murphymj25 for the confirmation.

Issue #807, opened 2020-12-17
grpc be closed when uninstall service target


qqwx1986:
there two problems
1. target connect will close while one target uninstall by method grpcConnectionPool::cleanup(),it's 5s interval check
I uninstall the service of the current target. Although I refused  new grpc request right now ,But I need to finish processing the grpc requests currently being processed before disconnecting

2. grpc_handle.go function  hasTarget compare by pointer(target address) to check target whether exist in route.Table, code main.go method watchBackend will NewTable every time while service refresh,so if one service had three target,if just one target uninstall from service ,other two target also will closed by cleanup() ,i think it need reuse old target connect

qqwx1986:
i resovled my problem by changed
--
```
func (p *grpcConnectionPool) cleanup() {
	for {
		p.lock.Lock()
		table := route.GetTable()
		for target, cs := range p.connections {
			if cs.GetState() == connectivity.Shutdown {
				delete(p.connections, target)
				continue
			}
			if !hasTarget(target, table) {
				if cs.GetState() != connectivity.Ready {//delete If not ready
					log.Println("[DEBUG] grpc: cleaning up connection to", target.URL.Host,cs.GetState())
					cs.Close()
					delete(p.connections, target)
				}
			}
		}
		p.lock.Unlock()
		time.Sleep(p.cleanupInterval)
	}
}

func hasTarget(target *route.Target, table route.Table) bool {
	for _, routes := range table {
		for _, r := range routes {
			for _, t := range r.Targets {
				if target.URL.Host == t.URL.Host {//compare by HOST
					return true
				}
			}
		}
	}
	return false
}
```

PR #808, opened 2020-12-21
fix all targets of route grpc connect will  be closed  when update route

qqwx1986:
fix issue #807

### DIFF
--- a/proxy/grpc_handler.go
+++ b/proxy/grpc_handler.go
@@ -275,9 +275,11 @@ func (p *grpcConnectionPool) cleanup() {
 			}
 
 			if !hasTarget(target, table) {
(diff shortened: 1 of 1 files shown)


cmvoicu:
Can anyone take a look at this PR?  The issue for droping gRPC streams on config update it's quite big, considering that the stream should be alive as long as the service is alive and is present in the registry!

nathanejohnson (commented):
After looking through the GRPC code I noticed several other issues that needed addressing.  I appreciate this PR, and I will make sure you get credit in the release notes for 1.5.16, but I believe this is the path forward: https://github.com/fabiolb/fabio/pull/816

PR #812, opened 2021-01-11
add option to constrain fabio instance to specific consul namespace

baabgai:
This helps to use fabio in a multi tenant environment by allowing to register a fabio instance in a specific consul namespace. All services that will be picked up with the specified tagprefix by this instance only come from the same consul namespace.


ketzacoatl:
@leprechau / @pschultz / @nathanejohnson, is this contribution you would want to see in fabio? It seems useful to those of us running multiple fabio instances (each with their own tagprefix), or looking to isolate groups of services/traffic.

pschultz:
Seems very useful to me. I just can't help with the review/testing, because we're not an enterprise customer.

nathanejohnson:
I can see this being useful, but like @pschultz we are not enterprise customers so I can't test or vet this PR.

ketzacoatl:
I'm not a gopher, so my review of the code would be meaningless. Would you be able to do a code review?

I could run tests to confirm  the expected behavior.

pschultz:
The code looks fine to me. It's a pretty obvious change. What I can't judge is if there are any missing places where the namespace has to be respected, but should become apparent with testing.

leprechau:
@baabgai are you using this patch in your environment? Can you confirm operation?

baabgai:
Hi @leprechau,
Yes we're using it at the moment in production and until we didn't experience any issues with namespace addition. To be fair I have to mention that we're still in the process of setting up a multi tenant cluster where this component is used and the current production load is still rather low.

tristanmorgan (approved):
Tested against latest Consul Enterprise and CE.

PR #816, opened 2021-02-23
fixes #807 - changes map for grpc connections to be a string key

### LINKED ISSUE #807: grpc be closed when uninstall service target

nathanejohnson:
This should be a more complete fix for #807 , inspired by https://github.com/fabiolb/fabio/pull/808

### DIFF
--- a/proxy/grpc_handler.go
+++ b/proxy/grpc_handler.go
@@ -54,8 +54,7 @@ func GetGRPCDirector(tlscfg *tls.Config) func(ctx context.Context, fullMethodNam
 			return ctx, nil, fmt.Errorf("error extracting metadata from request")
 		}
 
-		outCtx, _ := context.WithCancel(ctx)
-		outCtx = metadata.NewOutgoingContext(outCtx, md.Copy())
+		outCtx := metadata.NewOutgoingContext(ctx, md.Copy())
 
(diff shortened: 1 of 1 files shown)


nathanejohnson:
@qqwx1986 @cmvoicu can you take a look at this PR and see if this addresses your issue?

qqwx1986:
> @qqwx1986 @cmvoicu can you take a look at this PR and see if this addresses your issue?

it's addresses my issue,thanks

Vetrenik:
Is there some progress with merging this pull request?

yuriy-yarosh:
Yup, we need this...

qqwx1986:
> We're just waiting on confirmation from @qqwx1986. If someone else in a similar situation could test and confirm that would be appreciated.

i think this pr just fix one problem, and another i had reply up floor

PR #829, opened 2021-05-15
Allow Consul api to set default config for addr

### LINKED ISSUE #773: Unix socket support for registry.consul.addr?
My setup has most Consul communication happening over a Unix socket, and it looks like [`registry.consul.addr`](https://fabiolb.net/ref/registry.consul.addr/) doesn't work when given a file path or a `unix://` address. Is there a way to do this that I'm missing?

pingiun:
As can be seen in https://github.com/hashicorp/consul/blob/91ee7990ccf27ab7fccff746f8df0df65c3c583d/api/api.go#L390 the API also defaults to localhost:8500, however it can also use the environment variable CONSUL_HTTP_ADDR which is useful when fabio is run with nomad.

This change makes sure that the existing default stays working, but it now works better by default in more situations

Fixes #773

### DIFF
--- a/config/default.go
+++ b/config/default.go
@@ -51,8 +51,8 @@ var defaultConfig = &Config{
 	Registry: Registry{
 		Backend: "consul",
 		Consul: Consul{
-			Addr:            "localhost:8500",
-			Scheme:          "http",
+			Addr:            "",
+			Scheme:          "",
 			KVPath:          "/fabio/config",
 			NoRouteHTMLPath: "/fabio/noroute.html",
 			TagPrefix:       "urlprefix-",


nathanejohnson:
@pingiun I think this is a good change, sorry this has taken so long to get back to you.  But if you can get this rebased against master, i'll try to sneak this in with the next release which will likely happy tomorrow.

pingiun:
@nathanejohnson good to hear! I switched from fabio to traefik myself but I'm happy to help out

nathanejohnson:
So the config load tests are failing, I will have to postpone this for a bit when I have more time to dig in, but I believe it's in the parseScheme method for parsing the consul address.

PR #830, opened 2021-05-15
Add external authorization via http request

pingiun:
This is similar to [Envoys external authorization](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_authz/v3/ext_authz.proto), and can be used for example in combination with [Ory Oathkeeper](https://www.ory.sh/oathkeeper/docs/reference/api#access-control-decision-api).

### DIFF
--- a/auth/external.go
+++ b/auth/external.go
@@ -1,0 +1,63 @@
+package auth
+
+import (
+	"bytes"
+	"log"
+	"net/http"
+
+	"github.com/fabiolb/fabio/config"
+)
+
+type external struct {
+	endpoint          string
+	appendAuthHeaders []string
+	setAuthHeaders    []string
+}
+
+func newExternalAuth(cfg config.ExternalAuth) (AuthScheme, error) {
+	return &external{
+		endpoint:          cfg.Endpoint,
+		appendAuthHeaders: cfg.AppendAuthHeaders,
+		setAuthHeaders:    cfg.SetAuthHeaders,
+	}, nil
+}
+
+func (b *external) Authorized(request *http.Request, response http.ResponseWriter) AuthDecision {
+	client := &http.Client{CheckRedirect: func(req *http.Request, via []*http.Request) error { return http.ErrUseLastResponse }}
+
+	authRequest, err := http.NewRequest(request.Method, b.endpoint+request.URL.RequestURI(), bytes.NewReader(nil))
+	if err != nil {
+		log.Println("[ERROR] Can't make auth external request value:", err.Error())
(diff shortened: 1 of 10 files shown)


wfaler:
👍 for this PR!

auth/external.go
```diff
@@ -0,0 +1,63 @@
...
+func newExternalAuth(cfg config.ExternalAuth) (AuthScheme, error) {
+	return &external{
+		endpoint:          cfg.Endpoint,
+		appendAuthHeaders: cfg.AppendAuthHeaders,
+		setAuthHeaders:    cfg.SetAuthHeaders,
+	}, nil
+}
+
+func (b *external) Authorized(request *http.Request, response http.ResponseWriter) AuthDecision {
+	client := &http.Client{CheckRedirect: func(req *http.Request, via []*http.Request) error { return http.ErrUseLastResponse }}
```
nathanejohnson, line 26: This client should specify a timeout at least, and possibly a transport too depending on whether tls configurations would need to be overridden.

auth/external.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+	authRequest.Header = request.Header
+	authRequest.Host = request.Host
+
+	resp, err := client.Do(authRequest)
+	if err != nil {
+		log.Println("[ERROR] External request error:", err.Error())
+		return unauthorized()
+	}
+
```
nathanejohnson, line 41: resp.Body is never drained or closed.  if err is nil, body needs to be dealt with.

config/load.go
```diff
@@ -689,7 +690,23 @@ func parseAuthScheme(cfg map[string]string) (a AuthScheme, err error) {
...
 			a.Basic.Refresh = d
 		}
+	case "external":
+		a.External = ExternalAuth{
+			Endpoint:          cfg["endpoint"],
+			AppendAuthHeaders: strings.Split(cfg["append-auth-headers"], ","),
+			SetAuthHeaders:    strings.Split(cfg["set-auth-headers"], ","),
+		}
 
+		if strings.HasSuffix(a.External.Endpoint, "/") {
```
nathanejohnson, line 700: Curious to the reasoning for this restriction?

nathanejohnson (changes requested):
Overall I think this will be a great feature, just have a few tweaks before it can be merged.  Sorry this has taken so long to get back to.

marco-m:
Hello,
to avoid duplication, I suggest to double-check that this approach works also with `oauth2_proxy` (more details in #1006).

PR #836, opened 2021-06-25
Do not expand the source if mapping is missing

ABAC699:
Do not expand the source if the key is missing.


nathanejohnson:
Can you describe a use case for this?

818Nawaf (approved):

818Nawaf (approved):

Issue #369, opened 2017-10-10
Warn if fabio is run as root


magiconair:
Follow up from #195 which suggest that fabio should refuse to start as `root` unless you pass an `-insecure` flag. I'm still not fully convinced that users will just use the flag as workaround. Everybody has a right to shoot him/herself in the foot but I'm open for discussion.

sean-:
PostgreSQL does this and it’s been a useful safety precaution that has probably prevented many incidents.  That said, it’s also extremely irritating as an operator when diving into a problem as root.  Or even when I’m in a secured contained environment and there is only one process, but now I have to manage the creation and lookup of a user/UID.  In an ideal world I’d rather see a big obnoxious warning when the flag is used and a once-an-hour log entry nagging the user that the process is running as root (vs only at startup time).

mterron:
I stand by what I said in the original thread. It is the software responsibility to do the right thing, in this case refuse to run as root. If  you really really really want to, there's a cli switch that will allow you to do that. Let's make doing the right thing easy.
The last ~25 years has proven beyond doubt that "secure by default" is the only reasonable choice, as the large majority of users/operators WILL NOT take steps to secure their deployments and WILL ignore **"a big obnoxious warning when the flag is used and a once-an-hour log entry nagging the user that the process is running as root"**.

If you know what you are doing and want to run with decreased security for your environment, adding a cli switch that is clearly explained to you when you try to do something that's potentially unsafe is not a big deal. Same if you are building containers, adding a user on the build phase is trivial work and industry best practice.

magiconair:
@mterron I agree that the software should do the right thing but there is a valid case for running fabio as `root` inside a container if it is the only app. Static go binaries like fabio run from a `SCRATCH` container where there is no user management. Making this mandatory would mean to require a full-blown OS to support that which just increases the size and the attack surface. Also, containers provide the `root` abstraction for that reason, IMHO. You can be `root` but you're not `root` outside your container.

Also, if users will ignore a warning in the logs then they will just use the switch if is there. :)

OK, then the path forward is to refuse to run as `root` unless you specify `-insecure`. Issue a visible warning at startup and one WARN log message every hour.

mterron:
Just 1 comment. Container root (in linux) IS the host system root. There's
a new user mapping feature but not widely deployed afaik.

magiconair:
That's how I understand that but you'd have to break out of the container first, right? Isn't the point of containers to prevent that?

[...]

PR #851, opened 2021-11-07
Dockerfile: do not run as root

marco-m:
This switches off the warning

	************************************************************
	You are running fabio as root without the '-insecure' flag
	This will stop working with fabio 1.7!
	************************************************************

and makes all of us feel better :-)

See #369

NOTE In my limited tests it just works (using the tutorial https://learn.hashicorp.com/tutorials/nomad/load-balancing-fabio?in=nomad/load-balancing and stopping before section "Place Nomad client nodes behind AWS load balancer"), but since I am learning Nomad/Consul/Fabio, I might be missing something.

Fastest way to test:

    $ docker build -t <USER>/fabio:test-nonroot .
    $ docker push <USER>/fabio:test-nonroot

job spec:

    task "fabio" {
      driver = "docker"
      config {
        image        = "<USER>/fabio:test-nonroot"

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,5 +18,7 @@ RUN apk update && apk add --no-cache ca-certificates
 COPY --from=build /src/fabio /usr/bin
 ADD fabio.properties /etc/fabio/fabio.properties
 EXPOSE 9998 9999
+
+USER 1001
 ENTRYPOINT ["/usr/bin/fabio"]
 CMD ["-cfg", "/etc/fabio/fabio.properties"]


marco-m:
Hello @leprechau, pinging you here since you suggested this change in #369 and it seems that this project activity has slowed down, so I just wanted to be sure somebody is aware of this PR. Thanks!

nathanejohnson:
I am not going to merge this, though I do think there is validity to running as non-root.  The issue is running as user 1001 is not appropriate, this will actually be the second user created by default on many host operaating systems, and 1001 has no meaning inside of alpine as-is.  I may revisit this in the future - perhaps do the rube goldberg approach that mongodb uses where they have aa script as an entrypoint that tries to create a user on the fly, chown some files and run the mongod daemon as that user.  This makes more sense for a database where it will be writing to persistent volume mounts.  But at present, Traefik and Caddy both run as root in their official docker images, and so will fabio (for now).

PR #854, opened 2021-11-15
Update registry.backend.md

webmutation:
fix: dest to dst 
doc: added short description to the fields


dcarbone:
@webmutation Thank you for this PR, and apologies for taking so long to get back to you!  If you could, please make this same update in the [example fabio.properties](https://github.com/fabiolb/fabio/blob/master/fabio.properties#L625) file and I'll get it merged in :)

Issue #828, opened 2021-04-27
X-Forwarded-For doesn't seem to be working correctly


far-blue:
I'm struggling with the Forwarded and X-Forwarded-For headers on http traffic passing through Fabio.

Our Fabio instances sit behind a Cloudflare style WAF/LB setup and I can see, via tcpdump, that this setup is correctly passing X-Forwarded-For headers. However, traffic passing through Fabio seems to drop all that info and the only entry in the X-Forwarded-For (and Forwarded) headers is the IP address of the WAF/LB.

For instance:
With a client IP of 1.2.3.4
and a WAF IP of 5.6.7.8
and a Fabio IP of 9.9.9.9

I see the following using tcpdump from the WAF:

X-Forwarded-For: 1.2.3.4
X-Forwarded-Proto: https

I see the following using tcpdump coming from Fabio:

X-Forwarded-For: 5.6.7.8
X-Forwarded-Proto: https
Forwarded: for=5.6.7.8; proto=https; by=9.9.9.9; httpproto=http/2.0;

I would expect instead to see:

X-Forwarded-For: 5.6.7.8, 1.2.3.4
X-Forwarded-Proto: https
Forwarded: for=1.2.3.4; proto=https; by=9.9.9.9, by=5.6.7.8, httpproto=http/2.0;

I've tried leaving `proxy.header.clientip` blank and also setting it to 'X-Forwarded-For'. Looking at the go http source code I would have expected a blank value to equate to the behaviour I'm seeing and 'X-Forwarded-For' to correctly retain and extend the X-Forwarded-For header. But this is not what I'm seeing.

Anyone got any ideas?

leprechau:
So setting `proxy.header.clientip` to `X-Forwarded-For` or `X-Real-Ip` is going to be ignored.  That's a noop.

https://github.com/fabiolb/fabio/blob/81ed7aaf12ee494c676e0bbd830fd549ef0a4e1d/proxy/http_headers.go#L43-L50

The `Forwarded` behavior is expected. We don't pull from `X-Forwarded-For` to add to the `Forwarded` header. The `Forwarded` header is always set but will preserve existing information already present in a `Forwarded` header.

https://github.com/fabiolb/fabio/blob/81ed7aaf12ee494c676e0bbd830fd549ef0a4e1d/proxy/http_headers.go#L94-L114

The one condition where we might replace an upstream `X-Forwarded-For` header is in the case of a websocket connection.

[...]

PR #860, opened 2022-01-25
Fix x-forwarded-for header processing for ws connections

bn0ir:
Fix `x-forwarded-for` header processing logic for ws connections.
Instead of simply use remoteIP as `x-forwarded-for` header for backend, server add remoteIP to the end of `x-forwarded-for` header from client and send result to backend.

This PR possibly connected with issue https://github.com/fabiolb/fabio/issues/828

### DIFF
--- a/proxy/http_headers.go
+++ b/proxy/http_headers.go
@@ -58,7 +58,12 @@ func addHeaders(r *http.Request, cfg config.Proxy, stripPath string) error {
 	// http proxy which sets it.
 	ws := r.Header.Get("Upgrade") == "websocket"
 	if ws {
-		r.Header.Set("X-Forwarded-For", remoteIP)
+		targetHeader := []string{remoteIP}
+		sourceHeader := r.Header.Get("X-Forwarded-For")
+		if sourceHeader != "" {
+			targetHeader = append([]string{sourceHeader}, targetHeader...)
+		}
+		r.Header.Set("X-Forwarded-For", strings.Join(targetHeader, ","))
 	}
 
 	// Issue #133: Setting the X-Forwarded-Proto header to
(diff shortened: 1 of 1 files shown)


nathanejohnson:
@leprechau could you take a look at this one?  This rings of this old thread:

https://github.com/fabiolb/fabio/pull/449

proxy/http_headers.go
```diff
@@ -58,7 +58,12 @@ func addHeaders(r *http.Request, cfg config.Proxy, stripPath string) error {
 	// http proxy which sets it.
 	ws := r.Header.Get("Upgrade") == "websocket"
 	if ws {
-		r.Header.Set("X-Forwarded-For", remoteIP)
+		targetHeader := []string{remoteIP}
+		sourceHeader := r.Header.Get("X-Forwarded-For")
+		if sourceHeader != "" {
```
nathanejohnson, line 63: There are some subtle things going on here.  It would be good to mirror the functionality in the standard library:

https://github.com/golang/go/blob/master/src/net/http/httputil/reverseproxy.go#L296

The Get() call on the header only grabs the first element if there is more than one.

proxy/http_headers.go
```diff
@@ -58,7 +58,12 @@ func addHeaders(r *http.Request, cfg config.Proxy, stripPath string) error {
...
+			targetHeader = append([]string{sourceHeader}, targetHeader...)
+		}
+		r.Header.Set("X-Forwarded-For", strings.Join(targetHeader, ","))
```
nathanejohnson, line 66: Down here, these should be joined by 

    ", "

as it is done in the standdard library here:

https://github.com/golang/go/blob/master/src/net/http/httputil/reverseproxy.go#L303

nathanejohnson (changes requested):

leprechau:
Agreed, we should be trying to mirror the functionality in the standard library where possible. This also shouldn't have an impact on the ACL functionality.

nathanejohnson (approved):

ketzacoatl:
@nathanejohnson / @leprechau can this be released in a `v1.6.1`?

Issue #862, opened 2022-02-28
Experiencing 502's


aal89:
We use Fabio as a load balancer in a Nomad setup with Consul and some backend services (node.Http servers). With a couple of million of requests coming in each month we see that about 1500 requests are getting 502'ed. Percentage wise this is about a ~0.0007%, so a very low number. Nevertheless we'd like to solve this issue.

Reading through the issues we found a couple of similar looking ones. For example, see; https://github.com/fabiolb/fabio/issues/721
https://github.com/fabiolb/fabio/issues/716
___

We typically find these log lines in Fabio at the time of 502 errors:

```
http: proxy error: read tcp IP:33374->IP:24218: read: connection reset by peer
http: proxy error: EOF
```

These are indicators for a TCP RST packet, so the backend killed the connection already. We figured that it has to do something with incorrectly configured keepAlive's from both sides. See these two articles https://shuheikagawa.com/blog/2019/04/25/keep-alive-timeout/ & https://docs.apigee.com/api-platform/troubleshoot/runtime/502-bad-gateway for a better explanation. Even though the load balancers are different they do describe the same type of problem.

We tried a setup where we set the `proxy.keepalivetimeout` to 20s for Fabio and our backend service a `server.keepAliveTimeout` of 30s. This way the load balancer will always try to kill the connection and _not_ the backend service. However we found that all TCP connectings were still getting into a `TIME-WAIT` state on the backend service, indicating that the backend service initiated the closing of the socket. No matter what configuration we tried to set for Fabio, it didn't work. The backend service was always initiating closure.

Upon further investigation it seems that the node.Http server actually does some extra stuff when the `keepAliveTimeout` is being hit. It also destroys the socket (https://github.com/nodejs/node/blob/45b5ca810a16074e639157825c1aa2e90d60e9f6/lib/_http_server.js#L587), this behaviour is not found in Fabio when you set `proxy.keepalivetimeout`. It just keeps the socket there and eventually the backend service would kill the connection. Because it hits it's own `keepAliveTimeout`.

Additional testing did gave us some good results though. We found that we have to set `IdleConnTimeout` of the `http.Transport` to signal closure on the Fabio side after timeout. When configured this way (+ timeouts mentioned above) we noticed that the TCP connections on the backend services were no longer getting into a `TIME-WAIT` state, rather into a `CLOSE-WAIT` state, indicating that Fabio initiated closure. However, there's no `IdleConnTimeout` configurable in Fabio.

[...]

PR #863, opened 2022-02-28
Add IdleConnTimeout configurable for http transport

aal89:
Please see this issue for additional context https://github.com/fabiolb/fabio/issues/862.

I set a default value for `IdleConnTimeout` of 15s, because the `KeepAliveTimeout` default is 0s and `net.Dialer` will make it into a 15s default if it's 0s.

Thank you for reviewing.


nathanejohnson:

aal89:
Do we need to update some documentation too? This a quite a behavioural change as Fabio will now have a 15s keepalive for connections.

Issue #865, opened 2022-04-05
Consul Route updates very slow with large numbers of routes


ddreier:
We run several sets of Fabio instances, with the Consul backend, that register 20k and 60k routes. On the low end Fabio takes around a minute to process changes to the route table, and on the high end it's more like 5 to 6 minutes while Fabio is handling load.

I've been doing some experimenting and profiling of Fabio's code, and I believe that I might have identified a large time sink. Each time a new route is added into the route table, [the routes (per host) get sorted](https://github.com/fabiolb/fabio/blob/master/route/table.go#L174). We almost never do host-based routing, so nearly 100% of our routes all fall into the blank host route list.

I dumped one of our route tables into a file and wrote a test to load that file and then create a route table from those commands. Something like this:
```go
func TestNewTable(t *testing.T) {
	// read in a file with approximately 20k `route add` commands
	f, err := os.ReadFile("/path/to/some/fabio_routes.txt")
	if err != nil {...}

	b := bytes.NewBuffer(f)

	_, err := NewTable(b)
	if err != nil {...}
}
```

Without any code changes from release 1.5.15 this takes around 13 seconds on average.
```
=== RUN   TestNewTable
--- PASS: TestNewTable (13.42s)
PASS
```
CPU Profiling flame graph:
![image](https://user-images.githubusercontent.com/82290/161848495-2aaffa5b-a990-4fa2-a60d-2575f1c26146.png)

When I remove the Sort call from Table.addRoute and do it at the end of NewTable instead, the test completes in less than a second on average.
```
=== RUN   TestNewTable
--- PASS: TestNewTable (0.62s)
PASS
```
CPU Profiling flame graph:
![image](https://user-images.githubusercontent.com/82290/161848550-e5a509ed-16df-484a-aa41-94c9009ae30f.png)

So, that's a pretty significant improvement. I'm working on trying to get a more production-like test set up to measure the difference in update speed while Fabio is actually under load.

I'll have a pull request soon, and would really appreciate any feedback as well as consideration in merging and cutting a new release. Thanks!

PR #867, opened 2022-04-06
#865 Move the route table sort into NewTable so that it only happens once.

ddreier:
Addresses #865 where route table updates take multiple minutes with tens of thousands of routes.

### DIFF
--- a/route/table.go
+++ b/route/table.go
@@ -109,6 +109,12 @@ func NewTable(b *bytes.Buffer) (t Table, err error) {
 			return nil, err
 		}
 	}
+
+	// Sort the route table for each hostname
+	for _, h := range t {
+		sort.Sort(h)
+	}
+
 	return t, nil
 }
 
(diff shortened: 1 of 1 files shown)


nathanejohnson (approved):
This looks good, I'd like to get another set of eyes on this but it looks like addRoute is only called from NewTable and NewTableCustom , so I don't see this causing issues.

nathanejohnson:
So I have to ask...  why do you guys have so many routes? :)

ddreier:
> So I have to ask... why do you guys have so many routes? :)

We like to ask ourselves the same thing 😄 

Our Production Nomad clusters each run around 20k Allocations (we run mostly single-tenant services). Depending on the actual application, there could be as many as 6 or 8 routes registered per Alloc. So it ends up that we have Fabio instances for a couple of networks that have 60k routes.

nathanejohnson:

ddreier:
Thanks @nathanejohnson! Any hint on when the next version might be released? 😄

nathanejohnson:
@ddreier it was just released today.

PR #868, opened 2022-04-07
update go and alpine versions

Netlims:
There are many CVEs on the current alpine and golang version.
I have updated both and tested it on my local environment, it seems to have no effect on Fabio's functionality.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.15.5-alpine AS build
+FROM golang:1.17.8-alpine AS build
 
 ARG consul_version=1.8.2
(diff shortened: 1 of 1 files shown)


nathanejohnson:
This looks good.  If you sign the CLA I'll merge this ASAP.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.15.5-alpine AS build
+FROM golang:1.17.8-alpine AS build
```
nathanejohnson, line 2: This should be 

    FROM golang:1.18-alpine3.15 AS build

we should have always been specifying the alpine version so musl versions would match.

nathanejohnson (changes requested):

Netlims:
@nathanejohnson Good catch. I pushed the change and signed the CLA

nathanejohnson (approved):

PR #877, opened 2022-06-23
Documented preflight policy path (fixes  #876)

### LINKED ISSUE #876: Vault preflight check policy not documented
I just connected fabio to our vault and got stuck in a permission issue. Fabio always reported:
```
2022/06/23 07:17:03 [ERROR] cert: Cannot load certificates from secret/fabio/certs. vault: query mount path: Error making API request.

URL: GET https://cluster-vault-staging.fme.lan:8200/v1/sys/internal/ui/mounts/secret/fabio/certs
Code: 403. Errors:
```

The fun part is that this path is not documented as necessary policy entry in the [vault section](https://fabiolb.net/feature/certificate-stores/#vault). I checked the code and fount the function [kvPreflightVersionRequest](https://github.com/fabiolb/fabio/blob/81ed7aaf12ee494c676e0bbd830fd549ef0a4e1d/cert/vault_source.go#L145) as culprit, which should report the version of the kv secrets store. We use kv2. It works when the following section is added to the fabio policy in vault:

```
path "sys/internal/ui/mounts/secret/fabio/certs" {
  capabilities = ["read"]
}
```

This should be part of the documentation.

knorx:
Added section describing that newer versions of vault need the preflight check path to be readable.

### DIFF
--- a/docs/content/feature/certificate-stores.md
+++ b/docs/content/feature/certificate-stores.md
@@ -192,6 +192,11 @@ certificates are stored, for example:
       path "secret/fabio/cert/*" {
         capabilities = ["read"]
       }
+      
+      #to check for kv version in newer vault instances
+      path "sys/internal/ui/mounts/secret/fabio/certs" {
+        capabilities = ["read"]
+      }
 
 ##### Example
 


knorx:
Added documentation for token renewal, which was also missing.

nathanejohnson:
I set up a local vault instance running in dev mode, point my fabio instance to it and I was able to create the following minimal policy to support v2 keystore, with the preflight check.  It appears that it is unnecessary to use the full ui mount path and name this in the policy:

`path "secret/data/fabio/certs/*" {
  capabilities = ["read", "list"]
}

path "secret/metadata/fabio/certs"
{
  capabilities = ["list", "read"]
}`

The ability to self- renew is enabled in the default policies, though I do agree it would be useful to call it out in case some operator has removed this from the default policy.  This is the default policy on my local dev instance, which was just created:

`

# Allow tokens to look up their own properties
path "auth/token/lookup-self" {
    capabilities = ["read"]
}

# Allow tokens to renew themselves
path "auth/token/renew-self" {
    capabilities = ["update"]
}

# Allow tokens to revoke themselves
path "auth/token/revoke-self" {
    capabilities = ["update"]
}
`

If you could validate my findings, and update your documentation PR to include this, I'd like you to get credit for finding this.  Another tidbit, it's important to create the token as an orphan and with an explicit period, so that this will keep the token from bumping into the implicit max ttl which prevents the token from renewing after a time.

docs/content/feature/certificate-stores.md
```diff
@@ -192,6 +192,16 @@ certificates are stored, for example:
       path "secret/fabio/cert/*" {
         capabilities = ["read"]
       }
+      
+      # To check for kv version in newer vault instances
+      path "sys/internal/ui/mounts/secret/fabio/certs" {
+        capabilities = ["read"]
```
nathanejohnson, line 198: See the comments in the PR

nathanejohnson (changes requested):

PR #902, opened 2022-09-08
feat: Compatibility with vault-acme (fixes #900)

### LINKED ISSUE #900: Compatibility with acme-vault
As per https://github.com/remilapeyre/vault-acme/issues/34,
I would like to use fabio and acme-vault together to deal with letsencrypt certificates for loadbalancing and have vault deal with the certs through pki (e.g. acme-vault plugin).

However, the keys are different for when issuing certs. While this one is probably easy to get compatibility with, I wonder what is required to get these two wonderful projects talk more nicely with each other.

After all, I wonder how many people's needs this would satisfy.

Cheers,
 -- Fabian

xeroc:
First approach to making issuing of certificates compatible with the way vault-acme works.
Essentially, just allow to also load certificates from `cert` instead of `certificate`.

Since i am a Go novice, please let me know if there is a cleaner approach to this.

Also, am I supposed to direct PRs to `master`?

### DIFF
--- a/cert/vault_pki_source.go
+++ b/cert/vault_pki_source.go
@@ -69,9 +69,12 @@ func (s *VaultPKISource) Issue(commonName string) (*tls.Certificate, error) {
 
 	b, _ := json.Marshal(resp.Data)
 	var data struct {
+		// Try load from regular vault-pki
 		PrivateKey  string   `json:"private_key"`
 		Certificate string   `json:"certificate"`
 		CAChain     []string `json:"ca_chain"`
+		// Try to load data from acme-vault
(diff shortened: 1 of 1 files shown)


tristanmorgan:
I'm inclined to say the fix should really happen with the plugin as per [vault-acme#37](https://github.com/remilapeyre/vault-acme/pull/37). Also it looks like the plugin needs to get some attention (maintainers) with the last commit to HEAD being 3 years ago.

tristanmorgan:
I think a direct ACME implementation would probably serve Fabio users better than going via Vault. I'm going to close this one for now.

PR #907, opened 2022-09-22
Feature/source link

KTruesdellENA:
@Kamilcuk 

Added your changes and made it a configurable variable.  Can you review please?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,5 @@ fabio.sublime-*
 demo/cert/
 /pkg/
 dist/
+*.app
+*.hugo_build.lock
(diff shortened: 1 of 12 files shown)


Kamilcuk:
This is (with settings and such) way more than I could ever wish for. 

When hitting `make` it prints some `FAIL`, running `./fabio` results in error `panic: template: routes:86: function "RoutingTable" not defined`.

KTruesdellENA:
That issue should be fixed, i had pushed the initial changes before finding that bug, but it builds ok now and should be good.
We are holding on merging this initially right now while we hash out which "port" and "scheme" and what those should really be and where and when they should be used.  But test it and see if its what you are looking for and want and how this works for you.
Thank you

Kamilcuk:
I tested `ui.routingtable.source.linkenabled = true` and works fine for me.

nathanejohnson (approved):

Issue #912, opened 2022-11-18
Handle case where grpc backend may be deregistered from registry but is still handling traffic


nathanejohnson:
This is an attempt to gracefully shut down a connection to a grpc backend that gets deregistered from a registry, such as consul.  This is described as point 1 in #807 .  I'm not sure how common this will be, mostly a deregister event is caused by a failing health check, but this is easy enough to implement and shouldn't break anything so worth trying

PR #913, opened 2022-11-18
Updating grpc handler to gracefully close backend connections

nathanejohnson:
This should address part 1 of #807 and #912

update grpc docs


nathanejohnson:
@ qqwx1986  If you're still around curious to get your thoughts on this approach.

PR #914, opened 2022-11-20
[Docs] Fix wrong parameter name

KEANO89:
Starting fabio with `path` here results in an error which asks for a parameter `cert` so I guess that name changed since writing this doc


tristanmorgan (approved):
Correct, as documented at [Certificate Stores](https://fabiolb.net/feature/certificate-stores/#path)

PR #928, opened 2023-03-22
Consul registry performance improvements

ddreier:
Hi, it's me again trying to improve route update performance for our absurd number of routes. 😁

There's two major changes here:

The first is adding "filtering" so that only Consul Health Checks with the applicable Service Tag Prefix (and serf/maintenance checks) are considered during the rest of the process. In our Production environments this brings down the number of checks that Fabio has to look at from ~40k down to ~12k.

The second change is simplifying the `passingServices` function from an $O(n^5)$ process to $O(n^2)$. Each of the "helper" functions (`countChecks` `isAgentCritical` `isNodeInMaintenance` `isServiceInMaintenance`) did their own loop over each of the Consul Health Checks which added up in time very quickly.

Both of these changes drastically reduced the amount of time spent processing responses from Consul. In a test using real Consul data but not serving traffic we saw a reduction from around 90 seconds to less than 10 seconds. We've been testing Fabio with these changes in a non-production environment for about a week now, and have not detected any issues.
image
(The increase in `makeConfig` shown on the graph is due to how I measured it. It includes time waiting to write on the channel back to the main Fabio goroutine. This wasn't an issue before because processing the Consul data took longer than the time needed to build the actual route table (~10s) but the scales have now flipped. Building the Route Table takes longer than processing the data from Consul.)

I also had profiling running during my tests. Here's stock Fabio:
Stock Fabio CPU profile

After adding the Check "filtering":
Consul Health Check filtering CPU profile

After simplifying `passingServices`:
Filtering plus simplified passingServices

I'm very interested in any feedback about these changes and will do what I can to help satisfy any concerns. Thanks!

### DIFF
--- a/registry/consul/passing.go
+++ b/registry/consul/passing.go
@@ -7,28 +7,48 @@ import (
 	"github.com/hashicorp/consul/api"
 )
 
+// passingServices takes a list of Consul Health Checks and only returns ones where the overall health of
+// the Service Instance is passing. This includes the health of the Node that the Services Instance runs on.
 func passingServices(checks []*api.HealthCheck, status []string, strict bool) []*api.HealthCheck {
 	var p []*api.HealthCheck
+
+CHECKS:
 	for _, svc := range checks {
 		if !isServiceCheck(svc) {
 			continue
 		}
-		total, passing := countChecks(svc, checks, status)
+		var total, passing int
+
+		for _, c := range checks {
+			if svc.Node == c.Node {
+				if svc.ServiceID == c.ServiceID {
+					total++
+					if hasStatus(c, status) {
+						passing++
+					}
+				}
+				if c.CheckID == "serfHealth" && c.Status == "critical" {
+					log.Printf("[DEBUG] consul: Skipping service %q since agent on node %q is down: %s", c.ServiceID, c.Node, c.Output)
+					continue CHECKS
+				}
+				if c.CheckID == "_node_maintenance" {
+					log.Printf("[DEBUG] consul: Skipping service %q since node %q is in maintenance mode: %s", c.ServiceID, c.Node, c.Output)
+					continue CHECKS
+				}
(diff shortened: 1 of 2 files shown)


ddreier:
Hi @nathanejohnson and @KTruesdellENA, just wanted to check in and see if you have any feedback.

ketzacoatl:
@ddreier this looks awesome! I'd like to get it merged if I can get the access.
Is it feasible to add tests that would confirm these changes are valid and don't break anything?

aaronhurt:
> @ddreier this looks awesome! I'd like to get it merged if I can get the access. Is it feasible to add tests that would confirm these changes are valid and don't break anything?

Would definitely be nice to get some tests added, and I agree, this looks great.

tristanmorgan:
> Would definitely be nice to get some tests added, and I agree, this looks great.

I will have a crack at adding some tests next week if only to get this one merged.

Shackelford-Arden:
I know ddreier is fairly busy, but we've been running 30 or so instances of Fabio with this code with significantly improved performance :smile: We've been running this way for months now in production.

aaronhurt:
> I know ddreier is fairly busy, but we've been running 30 or so instances of Fabio with this code with significantly improved performance 😄 We've been running this way for months now in production.

Awesome, that's great to hear.

ddreier:
Unfortunately I'm not able to commit additional time right now. As @Shackelford-Arden mentioned, we've been running Fabio with these changes in production for over a year at this point. Our 30 or so Fabio instances handle hundreds of millions of public and internal requests per day.

The anecdote isn't the same as tests in code, for sure. But I hope it helps with confidence.

aaronhurt:
@ddreier thank you, and yes, it definitely helps.

tristanmorgan (approved):
Test coverage on passing.go is 100% and service.go is 0% (both unchanged from master), I've been running this code on my systems for at least 7 months without issue so I'm giving it a "LGTM".

PR #943, opened 2024-03-21
Trim leading and trailing spaces from service tags

logocomune:
This update modifies the 'routecmd' function to trim leading and trailing spaces from service tags. Corresponding tests have been added to verify correct functionality when space-prefixed or suffixed tags are encountered. These changes ensure consistency in how tags are processed and compared.

### DIFF
--- a/registry/consul/routecmd.go
+++ b/registry/consul/routecmd.go
@@ -26,6 +26,9 @@ type routecmd struct {
 func (r routecmd) build() []string {
 	var svctags, routetags []string
 	for _, t := range r.svc.ServiceTags {
+
+		t = strings.TrimSpace(t)
+
 		if strings.HasPrefix(t, r.prefix) {
 			routetags = append(routetags, t)
 		} else {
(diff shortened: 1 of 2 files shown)


registry/consul/routecmd.go
```diff
@@ -26,6 +26,9 @@ type routecmd struct {
 func (r routecmd) build() []string {
 	var svctags, routetags []string
 	for _, t := range r.svc.ServiceTags {
+
+		t = strings.TrimSpace(t)
```
ketzacoatl, line 30: @logocomune does this trim all spaces, or only leading/trailing spaces?

tristanmorgan: > TrimSpace returns a slice of the string s, with all leading and trailing white space removed, as defined by Unicode.

https://pkg.go.dev/strings#TrimSpace

tristanmorgan (approved):
Looks good. clean small change to fix a need.

PR #945, opened 2024-04-25
fix: close resp body

testwill:


tristanmorgan:
I think a better fix would be to use the following just after the `client.Do(req)` on line 48:
```
if resp != nil {
   defer resp.Body.Close()
}
```

aaronhurt:
bonus for catching the error in an anon func and logging it ...
```go
defer func() {
  if err := rest.Body.Close(); err != nil {
    log-error-here
  }()
```

tristanmorgan (approved):
our coverage for errors in this registry option isn't great but this should solve a potential resource leak.

PR #951, opened 2024-09-02
Remove vendored modules in favour of go mod.

tristanmorgan:
I believe this is the currently accepted best method for managing dependancies in Go. It should open the possibility for Dependabot and easier updates.


aaronhurt (approved):
There was a point in time where some vendored code was only contained in the `vendor` folder and was not available via a public repo. That looks to have changed. I agree this change reflects community best practice.

PR #952, opened 2024-09-05
Update go version, test binaries and package versions.

tristanmorgan:
This one is to address the list of upstream fixes by pulling in the latest of almost every dependency we have.

its going to sit for a while just to make sure it hasn't broken anything. The Tests pass but there could be areas that the coverage didn't uncover.


.github/workflows/build.yml
```diff
@@ -5,7 +5,7 @@ jobs:
     strategy:
       matrix:
         os: [ ubuntu-latest ]
-        go: [ '1.19.1', '1.18.6' ]
+        go: [ '1.23.0' ]
```
dcarbone, line 8: Here is a better update to this file:

```yaml
on: [ push, pull_request ]
name: Build
jobs:
  build:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout code
        uses: actions/checkout@v4

      - name: Install Go
        uses: actions/setup-go@v5
        with:
          go-version-file: 'go.mod'
          check-latest: true
          cache: true

      - name: Set Hosts
        run: |
          echo "127.0.0.1	example.com example2.com" | sudo tee -a /etc/hosts

      - name: Test
        run: |
          export PATH=$PATH:$HOME/bin:$HOME/go/bin
          make github
```

dcarbone (changes requested):
Overall looks good, but there is no reason to use a matrix anymore with the move to a singular version to test on, and may as well just modernize the whole thing.

.github/workflows/build.yml
tristanmorgan: Yes, and updating the actions too. (commit inbound)

Makefile
```diff
@@ -21,10 +21,10 @@ GOVERSION ?= $(shell go version | awk '{print $$3;}')
 GORELEASER ?= $(shell which goreleaser)
 
 # pin versions for CI builds
-CI_CONSUL_VERSION ?= 1.12.3
-CI_VAULT_VERSION ?= 1.11.0
+CI_CONSUL_VERSION ?= 1.18.2
```
dcarbone, line 24: Bump this to `1.19.2`, please.

dcarbone (changes requested):
Missed one consul version bump on my last pass, otherwise I think this is good.

@aaronhurt, any comments?

Makefile
tristanmorgan: done.

dcarbone (approved):

PR #963, opened 2025-01-14
README: remove mention to www.kijiji.it (decommissioned in 2022)

marco-m-pix4d:


tristanmorgan (approved):
Confirmed on the site it was taken down 31st May 2022.

Issue #941, opened 2024-02-21
Unable to load correct certificates if 1 invalid one is in consul k/v


finwo:
## Summary

When there's a single invalid certificate in the consul certificate store in use by fabio, all the other certificates in the store will not be refreshed.

Upon a fresh start of fabio (like a server reboot), this causes no single ssl endpoint to function anymore.

## Steps to reproduce

We're running fabio using the `fabiolb/fabio:latest` image from dockerhub on a nomad cluster as system job (running on every worker node).

First off, here's the used config loaded into /etc/fabio/fabio.properties within the container from a nomad template:

```
proxy.addr = :{{env "NOMAD_PORT_http"}};proto=http,:{{env "NOMAD_PORT_https"}}:proto=https;cs=cssl
ui.addr = :{{env "NOMAD_PORT_ui"}}
registry.consul.register.enabled = false
{{with nomadVar "nomad/jobs/fabio"}}
registry.consul.addr = {{.consul_addr}}
proxy.cs = cs=cssl;type=consul;cert=http://{{.consul_addr}}/v1/kv/certs
{{end}}
```

The certificates are loaded into `/certs` within consul as bundles, so `/certs/example.com.pem` contains the full chain including the private key.

To replicate the issue, insert a single certificate where the private key does not match the certificate. This will result in fabio not loading any certificates.

## Expected result

Fabio to load all valid certificates, and ignore/skips the invalid certificates, allowing domains where valid certificates are available for to continue to function.

## Actual result

Fabio refuses to load or update any certificate, causing all services it's loadbalancing to fail upon ssl initialization because it fails to find a valid certificate for the servername indicated by SNI.

tristanmorgan:
[func loadCertificates(...)](https://github.com/fabiolb/fabio/blob/master/cert/load.go#L107) tries to load all certs and if one fails, an error is returned.
The logs do reflect this and this is expected behaviour for the code as it stands now. should we change it is another question.

finwo:
For proof-of-concept setups having a single invalid one fail the entire batch makes sense, but for production setups having one invalid certificate not take down the entire updating or ingesting of all certificates "feels" unstable.

For proof-of-concept setups, the fail-fast approach makes sense, but for production setups having 1 update or invalid certificate be able to take away the ability to spin up new instances "feels" unstable. So even if both types of behaviors would be desirable depending on the context, having that be configurable would be a major improvement in my opinion.

[...]

PR #964, opened 2025-01-23
Report all certificate errors instead of stopping at the first. (#941)

tristanmorgan:
changed the loading of certificates process to report all errors found.

```
2025/01/23 22:44:34 [ERROR] cert: Failed to load certificates. 4 errors occurred:
	* cert: cannot load certificate nomad.service.consul-key.pem
	* cert: invalid certificate fabio.service.consul-cert.pem. tls: failed to find any PEM data in key input
	* cert: invalid certificate fabio.service.consul-key.pem. tls: failed to find any PEM data in key input
	* cert: cannot load certificate consul.service.consul-cert.pem
```

These errors reported were a result of deleting the nomad-cert, consul-key and uploading garbage to fabio-key.

### DIFF
--- a/cert/load.go
+++ b/cert/load.go
@@ -4,6 +4,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/pem"
+	"errors"
 	"fmt"
 	"io"
 	"log"
@@ -16,6 +17,7 @@ import (
 	"strings"
 )
 
+// MaxSize is the limit for individual certificate sizes.
 const MaxSize = 1 << 20 // 1MB
 
 func loadURL(listURL string) (pemBlocks map[string][]byte, err error) {
@@ -105,6 +107,7 @@ func loadPath(root string) (pemBlocks map[string][]byte, err error) {
 }
 
 func loadCertificates(pemBlocks map[string][]byte) ([]tls.Certificate, error) {
+	var errs []error
 	var n []string
 	x := map[string]tls.Certificate{}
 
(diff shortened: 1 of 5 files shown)


cert/load.go
```diff
@@ -14,8 +14,11 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
+
+	"github.com/hashicorp/go-multierror"
```
dcarbone, line 18: Unless one of the other maintainers has a counterpoint (@nathanejohnson or @aaronhurt ?), I would prefer we use the `errors` package for handling error concatenation.

The code can be updated to something like this:

```go
var errs []error

....
errs = append(errs, err)
...

return certs, errors.Join(errs...)
```

I'll defer to others if they prefer we use go-multierror, however.

dcarbone (changes requested):

cert/load.go
dcarbone: Note that `errors.Join` will return `nil` if there are no non-nil values in the provided slice, so you do not have to test for nil explicitly before appending to the `errs` slice.

tristanmorgan: if we do use the errors package I'll update this and [bgp/bgp_nonwindows.go](https://github.com/fabiolb/fabio/blob/5d113395f3d08045ebad183245fd69d9576dd317/bgp/bgp_nonwindows.go#L231) too.

dcarbone: I'm generally in favor of fewer deps whenever possible, even when they're as stable as hashi's go-multierror.

tristanmorgan: Trying the same manual test of deleting some keys, (using errors package instead) the error response is as follows:

```
2025/01/23 23:29:28 [ERROR] cert: Failed to load certificates. cert: cannot load certificate consul.service.consul-cert.pem
cert: invalid certificate fabio.service.consul-key.pem. tls: failed to find any PEM data in key input
cert: cannot load certificate nomad.service.consul-key.pem
cert: invalid certificate fabio.service.consul-cert.pem. tls: failed to find any PEM data in key input
```

dcarbone (approved):
lgtm, @nathanejohnson or @aaronhurt?

aaronhurt (approved):
lgtm as well, nice work

PR #965, opened 2025-01-31
update CI components

aleksraiden:
Update CI component version:

- Consul to 1.20.2
- Vault to 1.18.4
- Hugo to 0.142.0
- GoBGP to 3.33.0

PR #967, opened 2025-02-04
Update Hugo config to work with version bump in #965

tristanmorgan:


Makefile
```diff
@@ -185,7 +185,7 @@ github-pages:
 	mkdir -p ~/bin
 	tar -C ~/bin -zxf ~/hugo.tgz hugo
 	hugo version
-	(cd docs && hugo --verbose)
```
dcarbone, line 188: What does removing this flag do for us?

tristanmorgan: at some stage since hugo 0.101.0 the flag was removed and using it causes hugo to exit with an error code and help message.

tristanmorgan: see https://github.com/fabiolb/fabio/actions/runs/13146105854/job/36684473079

dcarbone: Excellent, thank you for the clarification!

dcarbone (approved):

PR #969, opened 2025-02-13
updating godeps

aleksraiden:
Update godeps to latest (as possible).

Currently, fabio deps cant be updated directly, like go get -u ./...  because we have two "points-of-pain": 

A first - armon/go-proxyproto and inetaf/tcpproxy cant be upgraded

Second - a part, working with zipkin and tracing. A newest version of tracing have a lot of changes in API, so we need a rewrite core logic into tracing module and test. A package, affected this: openzipkin-contrib/zipkin-go-opentracing, openzipkin/zipkin-go and apache/thrift.


tristanmorgan:
On a side note, I've been looking at replacing `github.com/armon/go-proxyproto` with `github.com/pires/go-proxyproto` but I have a troubling test to get working.

tristanmorgan (approved):
Tested and working well, Thanks for the contribution.

PR #971, opened 2025-02-14
Use a Go 1.24.0

aleksraiden:
Use a latest stable release of Golang - 1.24.0


tristanmorgan (approved):
Strictly speaking, the build pipeline will use the latest golang, and build an image with the Dockerfile-goreleaser. Also the go.mod version is more of a "this code with work with a minimum version of... ".

I'm accepting this PR mostly for the format string updates and having an updated go version in the Dockerfile will help local dev.

PR #974, opened 2025-02-24
fix: pollinterval to pollInterval

aleksraiden:
Fix a different case of option:

- registry.consul.pollinterval  => registry.consul.pollInterval
- registry.custom.pollinterval => registry.custom.pollInterval


tristanmorgan (changes requested):
Unfortunately, changing the case of this option would be a breaking change and its not the only one with less than ideal (letter) case.

PR #982, opened 2025-03-25
Document insensitive prefix matching in the list.

tristanmorgan:
Include case insensitive prefix matching in list of values.

### DIFF
--- a/docs/content/ref/proxy.matcher.md
+++ b/docs/content/ref/proxy.matcher.md
@@ -6,6 +6,7 @@ title: "proxy.matcher"
 `proxy.matcher` configures the path matching algorithm.
 
 * `prefix`: prefix matching
+* `iprefix`: case insensitive prefix matching
 * `glob`:  glob matching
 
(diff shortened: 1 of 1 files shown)


docs/content/ref/proxy.matcher.md
```diff
@@ -6,6 +6,7 @@ title: "proxy.matcher"
 `proxy.matcher` configures the path matching algorithm.
 
 * `prefix`: prefix matching
+* `iprefix`: insensitive prefix matching
```
marco-m, line 9: See also line 23 below which is somehow a duplicate. Maybe replace this line with line 23? Or in any case, I suggest "case insensitive", not only "insensitive".

tristanmorgan: This section fo the page should list all the options together. Updating it to case insensitive is a good suggestion that i will incorporate. Line 23 is part of a more detailed explanation so not really a duplicate.

dcarbone (approved):

PR #992, opened 2025-05-21
Using staticcheck to fix many issues.

tristanmorgan:
Fixes almost all the errors found by [staticcheck](https://github.com/dominikh/go-tools).


auth/auth_test.go
```diff
@@ -52,7 +52,7 @@ func TestLoadAuthSchemes(t *testing.T) {
 			t.Fatalf("could not create file on disk %s", err)
 		}
 
-		result, err := LoadAuthSchemes(map[string]config.AuthScheme{
+		result, _ := LoadAuthSchemes(map[string]config.AuthScheme{
```
dcarbone, line 55: I'm not super familiar with this code path, is ignoring errors OK here?

proxy/http_headers_test.go
```diff
@@ -495,5 +495,5 @@ func BenchmarkUint16Base16(b *testing.B) {
 			s = uint16base16(uint16(i))
 		}
 	})
-	_ = s == s // use the var to make go1.10 vet happy
+	fmt.Sprintf("hello %v", s) // use the var to make go1.10 vet happy
```
dcarbone, line 498: Why is this preferable to the old impl?

registry/custom/custom.go
```diff
@@ -46,13 +46,6 @@ func customRoutes(cfg *config.Custom, ch chan string) {
 	for {
 		log.Printf("[DEBUG] Custom Registry starting request %s \n", time.Now())
 		resp, err := client.Do(req)
-		if resp != nil {
-			defer func() {
-				if err := resp.Body.Close(); err != nil {
-					log.Printf("Error Can not close HTTP resp body - %s -%s \n", URL, err.Error())
-				}
-			}()
-		}
```
dcarbone, line 55: Why remove this defer?

dcarbone (changes requested):
Overall lgtm, just a couple questions.

registry/custom/custom.go
tristanmorgan: this one is easy, because its in an infinite loop, the function never exist so the defer is never triggered. that being said maybe I should get the loop to call a sub function that then exits properly.

auth/auth_test.go
tristanmorgan: the LoadAuthSchemes() can only throw an error if the type is not "basic".

proxy/http_headers_test.go
tristanmorgan: it was an attempt to avoid SA4000, but really just a non-functional change.

registry/custom/custom.go
tristanmorgan: resolved in the follow up commit.

proxy/http_headers_test.go
dcarbone: Mm, fair enough. Why don't we use a `t.Logf` statement instead, then call it good.

tristanmorgan: done.

dcarbone (approved):

PR #993, opened 2025-05-27
Bump go.mod pins to latest point release.

tristanmorgan:
Because the setup go action pull the specific version in the go.mod files it won't pull-latest ref [actions/setup-go/issues/561](https://github.com/actions/setup-go/issues/561).


tristanmorgan:
This also addresses CVE-2025-22871 , although Fabio doesn't do Chunked encoding its self and just passes the content direct from the backend.

dcarbone (approved):

PR #1003, opened 2025-09-12
docs: Fixes bad example for creating multiple basic authorization schemes

### LINKED ISSUE #929: proxy.auth does not work as specified in documentation
The following:

```
proxy.auth = name=default;type=basic;file=/passwd/default.htpasswd                                                                               
                     name=kamil;type=basic;file=/passwd/kamil.htpasswd    
```

Does not work and second entry is ignored.

The following does work:

```
proxy.auth = name=default;type=basic;file=/passwd/default.htpasswd,name=kamil;type=basic;file=/passwd/kamil.htpasswd    
```

It is parsed with https://github.com/fabiolb/fabio/blob/642e425cf664e23d257ca357b7b1bfd1e432683d/config/kvslice.go#L20

steffkelsey:
closes #929 

I traced the code path through parsing KV slices and the only way to create multiple maps is to separate with a comma. The examples in the docs will not work. This updates the docs.

### DIFF
--- a/docs/content/feature/authorization.md
+++ b/docs/content/feature/authorization.md
@@ -52,5 +52,5 @@ Supported htpasswd formats are detailed [here](https://github.com/tg123/go-htpas
     name=mybasicauth;type=basic;file=p/creds.htpasswd;refresh=30s
 
     # basic auth with multiple schemes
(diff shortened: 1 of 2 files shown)


tristanmorgan (approved):
Thanks for the contribution.

PR #1034, opened 2026-04-20
feat: add ui.path config option for serving UI/API behind a proxy

### LINKED ISSUE #323: Support different base path for UI/API
It would be nice to be able change the base path of the UI/API so that it can be served behind a proxy,
or even like this (fabio proxying itself):
`+ route add proxy /proxy http://192.168.1.1:9998/ tags "registrator" opts "strip=/proxy"`
without redirecting to /routes

maciej-lech:
## Summary

  Adds a new `ui.path` configuration option that sets a base path prefix for all UI and API routes. This allows fabio's admin interface to be served behind a reverse proxy at a sub-path.

  Closes #323

  ## Motivation

  Currently all admin UI and API routes are hardcoded to the root path (`/`). When fabio is placed behind a reverse proxy (or proxying itself) at a sub-path like `/fabio`, the UI breaks - it redirects to `/routes` instead of `/fabio/routes`, and all asset/API paths fail.

  ## Usage

  ```properties
  ui.path = /fabio
```

```sh
fabio -ui.path /fabio
```

  Then the UI is accessible at http://localhost:9998/fabio/routes

  When ui.path is not set, behavior is unchanged.


admin/server.go
```diff
@@ -29,47 +30,50 @@ func (s *Server) ListenAndServe(l config.Listen, tlscfg *tls.Config) error {
...
-	mux.Handle("/api/config", &api.ConfigHandler{Config: s.Cfg})
-	mux.Handle("/api/routes", &api.RoutesHandler{})
-	mux.Handle("/api/version", &api.VersionHandler{Version: s.Version})
-	mux.Handle("/routes", &ui.RoutesHandler{Color: s.Color, Title: s.Title, Version: s.Version, RoutingTable: s.Cfg.UI.RoutingTable})
-	mux.HandleFunc("/health", handleHealth)
+	mux.Handle(p+"/api/config", &api.ConfigHandler{Config: s.Cfg})
+	mux.Handle(p+"/api/routes", &api.RoutesHandler{})
+	mux.Handle(p+"/api/version", &api.VersionHandler{Version: s.Version})
+	mux.Handle(p+"/routes", &ui.RoutesHandler{Color: s.Color, Title: s.Title, Version: s.Version, Path: p, RoutingTable: s.Cfg.UI.RoutingTable})
+	mux.HandleFunc(p+"/health", handleHealth)
```
tristanmorgan: Due to how Fabio registers its own health-check with Consul, this line breaks the heath-check. one option would be to update the consul registration code, or just exclude the "/heath" path from the prefix (preferred).

P.S. can you fetch origin and rebase?

tristanmorgan (changes requested):

admin/server.go
maciej-lech: Ok, I tested everything with my staging Consul cluster, but I have `registry.consul.register.enabled = false` and manage the health-check directly with Nomad

tristanmorgan: Don't forget the test cases had the path added too.

admin/server_test.go
```diff
@@ -74,6 +75,37 @@ func TestAdminServerAccess(t *testing.T) {
...
+
+	roTestsWithPath := []test{
+		{"/fabio/api/manual", 403},
+		{"/fabio/api/paths", 403},
+		{"/fabio/api/config", 200},
+		{"/fabio/api/routes", 200},
+		{"/fabio/api/version", 200},
+		{"/fabio/manual", 403},
+		{"/fabio/routes", 200},
+		{"/fabio/health", 200},
```
tristanmorgan: this one..

admin/server.go
maciej-lech: Done

maciej-lech: Unfortunately now, I will not be able to use that health check via reverse proxy. It is not a problem for my case, but it makes this feature incomplete I guess

tristanmorgan: I think that the health check should not be exposed beyond the reverse proxy (its a Consul responsibility anyway).

tristanmorgan (approved):
Thank you for you contribution and patience.

PR #1042, opened 2026-05-25
Update github.com/osrg/gobgp to use v4 package.

tristanmorgan:
Making this one a draft because I don't have a good test environment to really exercise this part of the code.


tristanmorgan:
For anyone wanting to test a prebuilt image, I am sharing it via my docker hub [account](https://hub.docker.com/r/tristanmorgan/fabio). Caveats, this is where I experiment with things so it may have other things broken and generally follows whatever is in my fork of Fabio.

marco-m:
Hello @tristanmorgan,

FYI, upgrading to v4 would also fix vuln https://pkg.go.dev/vuln/GO-2026-4736
but I completely understand the 

> I don't have a good test environment to really exercise this part of the code.

me neither unfortunately.

(note that I found that fabio is vulnerable with `govulncheck`, which performs reachability analysis).

dcarbone (dismissed):

nathanejohnson:
I haven't had my lab set up for this in quite some time, but I was the original author of the feature so I'll kick the tires on this fork soon (today I hope)

bgp/logger.go
```diff
@@ -1,75 +1,102 @@
...
+		return level >= slog.LevelInfo
 	case "WARN":
-		return bgplog.WarnLevel
+		return level >= slog.LevelWarn
 	case "ERROR":
-		return bgplog.ErrorLevel
+		return level >= slog.LevelError
 	case "FATAL":
-		return bgplog.FatalLevel
+		return level >= slog.LevelError
```
marco-m: Consistency: in line 31 above we handle two case values in one (`case "TRACE", "DEBUG":`), while here ERROR and FATAL are separated ?

bgp/logger.go
```diff
@@ -1,75 +1,102 @@
...
-		fmt.Fprintf(&b, " %s=>%v", k, v)
+func (h *bgpLogHandler) Handle(_ context.Context, r slog.Record) error {
+	level := r.Level.String()
+	msg := r.Message
+
+	// Build the log message with attributes
+	var attrs string
+	r.Attrs(func(a slog.Attr) bool {
+		if attrs != "" {
+			attrs += " "
```
marco-m: This will allocate. For performance, consider instead `strings.Builder` (https://pkg.go.dev/strings#Builder) ?

bgp/bgp_nonwindows.go
```diff
@@ -276,40 +270,60 @@ func (bgph *BGPHandler) AddRoutes(ctx context.Context, routes []string) error {
...
+	}
+
+	// Add all paths at once
+	responses, err := bgph.server.AddPath(apiutil.AddPathRequest{
+		VRFID: "",
+		Paths: paths,
+	})
+	if err != nil {
+		return fmt.Errorf("bgp error adding paths: %w", err)
+	}
```
marco-m, line 342: To double-check my understanding:
In this case also if `len(errs) > 0` we only report `err` and return ?
While below we keep appending to `errs` ?

(mine is purely a consistency check, I am not familiar with the code)

bgp/logger.go
tristanmorgan: good catch, updating.

bgp/logger.go
tristanmorgan: done.

nathanejohnson (approved):

PR #1051, opened 2026-07-13
Fixes title in navbar not showing

froque:
Before (color red added in dev tools to show the problem):
image

After:
image


tristanmorgan (approved):
Thank you @froque for the fix. 

(Looks like I broke it in #998)

Issue #522, opened 2018-07-18
Fabio Manual Overrides Issue


satishviswanathan:
Fabio shows service in routing table even though the service is down and marked not healthy by consul

Scenario
********

We have a service running from 2 boxes registered with consul which does a health check and fabio shows the service in it's routing table with equal weight age. Now we have a manual override and reduced the traffic to one of the server to 35% using the following command.

route weight RFGreen /green weight .35 tags "/green2"

Now fabio shows 65% traffic to 1 server and 35% to other. Next we stop the server which is receiving 35% of the traffic and consul shows the service as not healthy but fabio still shows the route in it's routing table.

Is there a way we can have this route removed from the routing table when the service is not healthy?

momania:
That problem is related to this issue: https://github.com/fabiolb/fabio/issues/224 
As soon as there is an error in the manual override (in this case the tag can't be matched anymore), the auto updating of the routing table stops.

Issue #282, opened 2017-05-02
entering a bad route breaks route updates


killcity:
Noticed a scenario where entering a route (weight in this case) broke future route updates. Can anything be done to have reject a bad route if not in the correct format? I know the easy answer is: "Dont enter bad routes", but when teams grow, more "safety net" features become necessary.

Here is an example of a route that broke updates:

```route weight myapp http://myapp.com/ weight 0.55 tags "gimme_55percent"```

magiconair:
Is this a weight for a non-existing service? See #224

But in general I agree. There needs to be better feedback.

killcity:
This was for an existing service. Would it be relatively easy to enforce better strictness for the route format? Maybe fire off something in the logs showing:

```Bad route format: route weight xxxx xxxxxxx, route table not modifed```

killcity:
When weighting using `src` and `tags`, does every tag need to be specified or will a single tag allow the upstreams to be load balanced.

For instance, if I want to balance to two machines, does this route work to make sure anything with `ferrari` get 90% traffic or do I need to explicitly list both tags?

server1: `www, ferrari`
server2: `www, yugo`

`route weight mysite weight 0.9 tags "ferrari"`

Spicykoala:
I was actually just about to put this issue in. We've been using some automation to handle route overrides and cleanup directly in consul /fabio/config. The ideal functionality would be to log the bad route command and skip it (but not ignore the rest). Are there downsides to this behavior?

We made adjustments to the cleanup processes so that orphaned routes shouldn't happen, but it does scare me a little for when we're managing a large number of services. Also, there's a scenario where a service goes entirely offline and is pulled from the route table completely - causing a once valid route weight to become invalid and cause updates issues prior to cleanup.

I'd love to hear thoughts and/or suggestions on this.

magiconair:
I'm going to make fabio ignore syntax errors in routes and just log about them. Blocking the update of the routing table isn't a good approach.

spicykoala:
I was wondering what the status of this issue was? It's been causing some stability issues for large numbers of servers. If the route table stops updating, any scaling events or anything that happen during that state, the new/old nodes registered in consul aren't updated in the route table and everything starts dying. 

This seems to happen if a manual route is added that points to a non-existent service. (E.g., adding a new service record, where no docker containers have come online yet and haven't had a chance to create consul entries).

murphymj25:
[...]

Issue #272, opened 2017-04-26
Fabio do not listen on proxy.addr when additinal route has no target match


sielaq:
Hi,

We have an issue.
After restarting Fabio, when having some "future weight for routes" in `registry.consul.kvpath` (consul), fabio is not willing to start listening on `proxy.addr`
and complain log says:
`[WARN] route: no target match`

version affected 1.4.x  (tested on 1.4.0 and 1.4.3)
with the same additional routes configuration in kv 
fabio version 1.3.x works fine:
start listening on `proxy.addr port`, and then complains.

Desired state:
start listening even if there are routes that do not match.

example of k/v in consul:
```
route weight foo foo.example.com/ weight 1.0 tags "fooA"
route weight foo foo.example.com:8080/ weight 1.0 tags "fooA"
route weight foo foo.example.com/ weight 0.0 tags "fooB"
route weight foo foo.example.com:8080/ weight 0.0 tags "fooB"
```

magiconair:
Sounds like #224

sielaq:
any news ?

magiconair:
@sielaq I'm falling a bit behind :( I'll try to catch up on open issues ...

PR #1053, opened 2026-07-14
Ignore invalid route weight commands

### LINKED ISSUE #224: Routing table not updated when rule references non-existent service

froque:
Previously, this would fail to construct a table and would not update the routes.

Fixes #224, #522, #272 and #282


tristanmorgan (approved):
Thanks again.

PR #1054, opened 2026-07-14
Fixes values in label target for metrics

froque:
The values array contains the labels and its values, to we should skip every other entry to get to the actual value.

❯ consul agent -dev
❯ go run ./demo/server/server.go -addr 127.0.0.1:5000  -prefix /foo -proto http
❯ ./fabio --metrics.prefix prefix1_fabio --metrics.target label 
❯ curl http://localhost:9999/foo/bar

Before change:
```
prefix1_fabio.requests:0|ms|#service:service
prefix1_fabio.route:0|ms|#service:service,host:server,path:host,target:
prefix1_fabio.http.status:0|ms|#code:code,service:404
```

After change:
```
prefix1_fabio.requests:1|ms|#service:server
prefix1_fabio.route:1|ms|#service:server,host:,path:/foo,target:http://127.0.0.1:5000/
prefix1_fabio.http.status:1|ms|#code:404,service:server
```

### DIFF
--- a/metrics/names.go
+++ b/metrics/names.go
@@ -62,8 +62,8 @@ func Labels(labels, values []string, stringsprefix, fieldsep, recsep string) str
 		}
 		_, _ = b.WriteString(labels[i])
 		_, _ = b.WriteString(fieldsep)
-		if i < len(values) {
-			_, _ = b.WriteString(values[i])
+		if (i*2)+1 < len(values) {
+			_, _ = b.WriteString(values[(i*2)+1])
 		}
 	}
 	return b.String()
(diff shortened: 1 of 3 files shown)


tristanmorgan:
I am trying to replicate the change in output but not sure what configuration it requires. for example I'm running Fabio with the following command line:

    ./fabio --metrics.target "stdout" --registry.consul.register.tags "urlprefix-/,admin"

Does this PR depend on #1059 ?

froque:
yes, #1059 also changed the output. 

I have rebased the branch and updated the description

tristanmorgan (approved):

Issue #1006, opened 2025-09-25
Support for authentication middleware (eg: oauth2-proxy) ?


marco-m-pix4d:
Hello,

probably best way to explain is to have a look at the [Architecture diagram of oauth2-proxy](https://oauth2-proxy.github.io/oauth2-proxy/#architecture) and see the image on the right: "Oauth 2 proxy as authentication middleware", although the feature is not exclusive to oauth2-proxy.

Both nginx and Caddy support this concept. Nginx calls it [auth_request](https://nginx.org/en/docs/http/ngx_http_auth_request_module.html); Caddy calls it [forward_auth](https://caddyserver.com/docs/caddyfile/directives/forward_auth).

Only mention of this feature I found is a ticket from 2016: https://github.com/fabiolb/fabio/issues/64.

Are there any plans to support this feature? If no, I can totally understand :-)

Thanks!

tristanmorgan:
Hi @marco-m-pix4d,

There are no current plans to implement authentication middleware but it is an interesting concept. The left diagram is interesting, I will have a test of that in my own home lab. As for the middleware style on the right, if it could be done in a way fitting with Fabio's lightweight style it would be a welcome addition to the project.

Thanks,

T.

marco-m-pix4d:
Thanks for the feedback @tristanmorgan, appreciated.

marco-m-pix4d:
Hello @tristanmorgan, I am coming back to this issue and I am confused by the resolution as **completed**.

If I look at the diffs between v1.6.9 and v1.6.10 https://github.com/fabiolb/fabio/compare/v1.6.9...v1.6.10, the CHANGELOG for **v1.6.10** does say:
```
## [v1.6.10](https://github.com/fabiolb/fabio/tree/v1.6.10) (2025-11-24)

[Full Changelog](https://github.com/fabiolb/fabio/compare/v1.6.9...v1.6.10)

**Implemented enhancements:**
 
- Support for authentication middleware \(eg: oauth2-proxy\) ? [\#1006](https://github.com/fabiolb/fabio/issues/1006)
```

but then looking at the diffs between v1.6.10 and v1.6.11 (https://github.com/fabiolb/fabio/compare/v1.6.10...v1.6.11), the CHANGELOG, note still for for **v1.6.10** has been modified and became:
```
## [v1.6.10](https://github.com/fabiolb/fabio/tree/v1.6.10) (2025-11-24)

[...]

PR #1071, opened 2026-08-24
tidy up and increase test coverage for basic auth

marco-m:
Hello @tristanmorgan,

while working on https://github.com/fabiolb/fabio/issues/1006 I did a detour on basic auth and found some possibilities to tidy up the tests.

Best reviewed commit-per-commit.

- auth/basic_test: booleans can be compared directly
- auth/basic_test: strings can be compared directly
- auth/basic_test: use self-describing names
- auth/basic_test: make test helpers idiomatic
- auth/basic_test: simplifications
- auth/basic_test: actually test the logs

### DIFF
--- a/auth/basic.go
+++ b/auth/basic.go
@@ -19,7 +19,7 @@ type basic struct {
 
 func newBasicAuth(cfg config.BasicAuth) (AuthScheme, error) {
 	bad := func(err error) {
-		log.Println("[WARN] Error processing a line in an htpasswd file:", err)
+		log.Println("[WARN] Error processing htpasswd file:", err)
 	}
 
 	secrets, err := htpasswd.New(cfg.File, htpasswd.DefaultSystems, bad)
--- a/auth/basic_test.go
+++ b/auth/basic_test.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"net/http"
 	"os"
-	"reflect"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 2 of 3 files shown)


auth/basic.go
```diff
@@ -19,7 +19,7 @@ type basic struct {
 
 func newBasicAuth(cfg config.BasicAuth) (AuthScheme, error) {
 	bad := func(err error) {
-		log.Println("[WARN] Error processing a line in an htpasswd file:", err)
+		log.Println("[WARN] Error processing htpasswd file:", err)
```
marco-m, line 22: I shorted this log because it repeats the same information in the log.

auth/basic_test.go
```diff
@@ -36,94 +37,90 @@ func (rw *responseWriter) WriteHeader(statusCode int) {
 	rw.code = statusCode
 }
 
-func createBasicAuthFile(contents string, t *testing.T) (string, error) {
+func createBasicAuthFile(t *testing.T, contents string) string {
```
marco-m, line 40: Test helpers by convention take as first parameter the *testing.T.

The t.Helper() makes it so that the test failure will come from the calling place of createBasicAuthFile(), which is normally more useful.

marco-m: Test helpers should not return errors, they should assert directly.

tristanmorgan (approved):
Thank you for your contribution. Nice work.

PR #1072, opened 2026-08-26
auth: tests: reduce runtime from 2.3 sec to 0.3 sec thanks to syncttest

marco-m:
Hello @tristanmorgan,

This PR reduces the runtime of the auth tests from 2.3 sec to 0.3 sec thanks to `syncttest` :-)

On the other hand, due to `synctest` technicalities, I had to slightly modify also the production code, to allow the test code to require termination of the goroutine spawned by `newBasicAuth()`.

If you are not comfortable with this, I can understand.

Unfortunately the diff is confusing. Better reviewed commit-per-commit.

References:
- https://go.dev/blog/testing-time
- https://antonz.org/go-concurrency/testing/

### DIFF
--- a/auth/basic.go
+++ b/auth/basic.go
@@ -11,12 +11,15 @@ import (
 	"github.com/tg123/go-htpasswd"
 )
 
-// basic is an implementation of AuthScheme
+// basic implements Basic HTTP Authentication.
+// It satisfies interface [AuthScheme].
 type basic struct {
 	secrets *htpasswd.File
 	realm   string
 }
 
+// newBasicAuth creates a [basic] authentication from cfg.
+// It might spawn a forever-running goroutine to periodically refresh the htpassd file.
 func newBasicAuth(cfg config.BasicAuth) (AuthScheme, error) {
 	bad := func(err error) {
 		log.Println("[WARN] Error processing htpasswd file:", err)
@@ -74,6 +77,7 @@ func newBasicAuth(cfg config.BasicAuth) (AuthScheme, error) {
 	}, nil
 }
 
+// Authorized returns whether request satisfies the authorization scheme.
 func (b *basic) Authorized(request *http.Request, response http.ResponseWriter) bool {
 	user, password, ok := request.BasicAuth()
 
--- a/auth/basic.go
+++ b/auth/basic.go
@@ -16,6 +16,8 @@ import (
 type basic struct {
 	secrets *htpasswd.File
 	realm   string
+	//
+	done chan struct{} // USE ONLY FOR TESTING.
 }
 
 // newBasicAuth creates a [basic] authentication from cfg.
@@ -31,6 +33,7 @@ func newBasicAuth(cfg config.BasicAuth) (*basic, error) {
 	basicAuth := &basic{
 		secrets: secrets,
 		realm:   cfg.Realm,
+		done:    make(chan struct{}),
 	}
 	if cfg.Refresh == 0 {
 		// In this case the htpassd file will not be reloaded, we are done.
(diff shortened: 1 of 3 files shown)


auth/basic_test.go
```diff
@@ -166,37 +167,41 @@ func TestBasic_Authorised(t *testing.T) {
...
-	if err != nil {
-		t.Error(err)
-	}
-
-	r := &http.Request{
-		Header: http.Header{
-			"Authorization": basicAuthHeader("foo", "bar"),
-		},
-	}
+	synctest.Test(t, func(t *testing.T) {
```
marco-m, line 170: This creates a synctest "bubble". Inside it, the time is faked.

To work, synctest needs all goroutine started within the bubble to terminate before the bubble itself terminates.

This is why I needed to add the stop() method to the prod code. I will add a comment there also.

auth/basic.go
```diff
@@ -64,16 +79,18 @@ func newBasicAuth(cfg config.BasicAuth) (AuthScheme, error) {
 						log.Println("[WARN] Error reloading htpasswd file:", err)
 					}
 				}
+			// A ticker can be stopped but not closed, so we need another channel.
+			// Used only for testing.
+			case <-basicAuth.done:
+				return
```
marco-m, line 85: We need this to make the goroutine terminate.
See the comment in the tests below to fully understand.

auth/basic_test.go
```diff
@@ -166,37 +167,41 @@ func TestBasic_Authorised(t *testing.T) {
...
+		// call to Authorized() to fail, as documented.
+		if err := os.Remove(filename); err != nil {
+			t.Fatalf("removing htpasswd file: %s", err)
+		}
 
-	t.Run("should not authorize after removing htpasswd file", func(t *testing.T) {
+		// Wait to ensure that the htpasswd file refresh happened.
+		// This happens within a syntest.Test(), so it will take zero time.
+		time.Sleep(2 * time.Second)
+		basicAuth.stop() // Technicality: synctest needs the goroutine to terminate.
```
marco-m: mhhhhh now that I think more about it, as-is this is a race :-/

tristanmorgan: If its not ready, please convert this PR to a draft.

marco-m: Converted to draft. This race seems tricky to me. Will ping when I have a satisfying idea.

marco-m: @tristanmorgan Ready for review.

In the previous commit, `basic.stop()` was closing the done channel,
but closing a channel is async, so `stop()` would return immediately to the test, with a possibility for the goroutine to still exist.

Now `basic.stop()` blocks sending on the channel until the receiving
goroutine has read the message. The race is potentially still there but
I think impractical. I ran `go test -count=10000 ./auth` with no flakes.

auth/basic.go
```diff
@@ -84,3 +101,13 @@ func (b *basic) Authorized(request *http.Request, response http.ResponseWriter)
 
 	return b.secrets.Match(user, password)
 }
+
+// stop terminates the goroutine that periodically refreshes the htpassd file.
+// stop will block until the goroutine receives the message.
+// If the goroutine is not running, Stop will block forever.
+// USE ONLY FOR TESTING.
+func (ba *basic) stop() {
```
tristanmorgan: mostly looking pretty good. This stop() function is only used it test (as intended) so would you consider moving it to the auth/basic_test.go file. That way we can ensure it is not included in the build. (and deadcode is happier.)

P.S. do you mind doing a quick rebase off master?

marco-m: Ah.
At the end, I removed the stop() entirely and sent directly from the test. Let me know your opinion.
I also rebased.

auth/basic_test.go
```diff
@@ -166,37 +167,46 @@ func TestBasic_Authorised(t *testing.T) {
...
 
-	t.Run("should not authorize after removing htpasswd file", func(t *testing.T) {
+		// Wait to ensure that the htpasswd file refresh happened.
+		// This happens within a syntest.Test(), so it will take zero time.
+		time.Sleep(2 * time.Second)
+		// Before exiting from the synctest bubble, we must terminate the basicAuth
+		// goroutine.
+		// We don't close the channel because closing is async. The send instead, since
+		// the channel is unbuffered, ensures that the test waits for the goroutine
+		// to receive and terminate.
```
marco-m, line 206: If it seems obvious, I can delete this part.

tristanmorgan (approved):
looking great, clear and logical.

PR #1081, opened 2026-09-04
Question about X-Forwarded-Proto

### LINKED ISSUE #957: Question about X-Forwarded-Proto
Hey all, im using nginx to terminate tls and sending through nginx all the possible forwarded-for headers

From the nginx i proxy pass to fabio on http,

My question is, will fabio change the forwarded-for-proto to http? Since it received the tarffic through http?

Since keycloak expects it to be found as https... 
(It has the option to recognize that tls was terminated at reverse proxy level, but im afraid fabio might be hondering it?)

Thanks in advance!

victorwon2001:
## Summary
Preserve the protocol already resolved by `addHeaders` after `ReverseProxy.SetXForwarded`, so TLS terminated by an upstream proxy can continue to reach backends as `X-Forwarded-Proto: https` even when Fabio receives the connection over HTTP.

## Changes
- Reapply the resolved inbound `X-Forwarded-Proto` after `SetXForwarded` while retaining the existing X-Forwarded-For append and X-Forwarded-Host behavior.
- Add integration coverage for forwarded HTTPS and the normal direct HTTP default.

## Testing
The focused regression test and the full proxy package test suite pass.

Fixes #957

### DIFF
--- a/proxy/http_handler.go
+++ b/proxy/http_handler.go
@@ -32,9 +32,13 @@ func newHTTPProxy(target *url.URL, tr http.RoundTripper, flush time.Duration) ht
 				req.Out.Header.Set("X-Forwarded-For", xff)
 			}
 
-			// SetXForwarded will handle X-Forwarded-For (append), X-Forwarded-Host, and X-Forwarded-Proto
-			// Other headers (X-Forwarded-Port, X-Forwarded-Prefix, Forwarded) are already set by addHeaders()
+			// SetXForwarded handles X-Forwarded-For (append) and X-Forwarded-Host,
+			// but derives X-Forwarded-Proto from the direct connection. Preserve the
+			// protocol already resolved by addHeaders(), including one from an upstream proxy.
 			req.SetXForwarded()
+			if xfp := req.In.Header.Get("X-Forwarded-Proto"); xfp != "" {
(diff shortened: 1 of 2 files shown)


proxy/http_handler.go
```diff
@@ -32,9 +32,13 @@ func newHTTPProxy(target *url.URL, tr http.RoundTripper, flush time.Duration) ht
...
 
-			// SetXForwarded will handle X-Forwarded-For (append), X-Forwarded-Host, and X-Forwarded-Proto
-			// Other headers (X-Forwarded-Port, X-Forwarded-Prefix, Forwarded) are already set by addHeaders()
+			// SetXForwarded handles X-Forwarded-For (append) and X-Forwarded-Host,
+			// but derives X-Forwarded-Proto from the direct connection. Preserve the
+			// protocol already resolved by addHeaders(), including one from an upstream proxy.
 			req.SetXForwarded()
+			if xfp := req.In.Header.Get("X-Forwarded-Proto"); xfp != "" {
+				req.Out.Header.Set("X-Forwarded-Proto", xfp)
+			}
```
tristanmorgan, line 41: I have been testing builds off HEAD and this is not needed.
to test in your own environment use curl with the `--haproxy-protocol` flag.

tristanmorgan (changes requested):
I don't think we need this.

PR #1085, opened 2026-09-08
registry/custom: tests: shave off 3 seconds

marco-m:
Before

    go test ./registry/custom -count=1
    ok      github.com/fabiolb/fabio/registry/custom        3.379s

After

    go test ./registry/custom -count=1
    ok      github.com/fabiolb/fabio/registry/custom        0.375s

As a reminder, since `go test ./...` runs each package test in parallel, this does not mean that the runtime of the _whole_ suite is reduced by 3s.

### DIFF
--- a/registry/custom/custom_test.go
+++ b/registry/custom/custom_test.go
@@ -3,16 +3,15 @@ package custom
 import (
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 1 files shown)


tristanmorgan (approved):
All logical steps and corrects the usage of the error condition.
Thank you.

PR #1094, opened 2026-09-17
addHeaders tests: tidy up

marco-m:
This is a refactoring-only (no behavior change) PR that, at least in my opinion :-), tidies up the addHeaders tests.

Rationale: while working on my proposal for #1089, I went looking for the existing tests regarding headers manipulation, and found some aspects that could be tidied up and be made more readable.

Best reviewed commit-per-commit.

Issue #1089, opened 2026-09-14
changes to X-Forwarded-Proto between v1.7 and v1.8


marco-m:
We deploy Fabio behind an AWS ELB. The ELB does TLS termination:
```
   Internet     |   VPC
                |
        HTTPS   |  HTTP        HTTP
browser =====> ELB ----> Fabio ---> service
```

The ELB sets correctly, from its point of view, header field `X-Forwarded-Proto` to `https`.

Fabio v1.7 (and before) was, I think, just copying the value it found of `X-Forwarded-Proto`, so in this case `https`.
From my understanding this was a security vulnerability.

Fabio v.1.8 is instead setting `X-Forwarded-Proto` to what it is actually seeing, so in the deployment above, it is setting it "back" to `http`.

In the changes from v1.7 to v1.8, I saw that function `newHTTPProxy()` now returns an `httputil.ReverseProxy `with the safer `Rewrite` as opposed to the unsafe `Director`: https://github.com/fabiolb/fabio/blob/2aad4553c7f96011ee39db91896a691ee745cb84/proxy/http_handler.go#L17

I still have to do more tests, but I think that this explains the changes in the handling of `X-Forwarded-Proto`.

As far as I understand, the behavior of Fabio v1.8.x is the correct one.

And now to my question:

There are many HTTP services that require `X-Forwarded-Proto` to be `https` in order to function. This is how actually we discovered the change I am describing: from an error from one of our services.

What I think I would like to be able to do is the following: somehow tell Fabio: yes, you are forwarding an HTTP (and not HTTPS) connection, BUT you are deployed behind a TLS terminating LB, you can trust the `X-Forwarded-XXX` header fields that you see.

Example: `oauth2_proxy` has a setting `--reverse-proxy` to tell it: you are in a chain you can trust. Doc says:  

> When `--reverse-proxy` is enabled, configure `--trusted-proxy-ip` to the IPs or CIDR ranges of the reverse proxies that are allowed to send `X-Forwarded-*` headers. If you leave it unset, OAuth2 Proxy currently trusts all source IPs for backwards compatibility, which means a client that can reach OAuth2 Proxy directly may be able to spoof forwarded headers.  

[https://oauth2-proxy.github.io/oauth2-proxy/configuration/overview?_highlight=reverse_proxy#proxy-options](https://oauth2-proxy.github.io/oauth2-proxy/configuration/overview?_highlight=reverse_proxy#proxy-options)

What do you think ?

[...]

PR #1096, opened 2026-09-18
headers integration tests: tidy up

marco-m:
This PR is in the same spirit of #1094, but while #1094 was about the unit tests for the headers, this one is about the integration tests for the headers.

This is a refactoring-only (no behavior change) PR that, at least in my opinion :-), tidies up the integration tests around which headers are overwritten and which headers are preserved.

Rationale: while working on my proposal for https://github.com/fabiolb/fabio/issues/1089, I went looking for the existing tests regarding headers manipulation, and found some aspects that could be tidied up and be made more readable.

Best reviewed commit-per-commit.

Note: While working on this PR, and comparing the unit tests with the integration tests, I also realized that the unit tests are misleading (the results are not meaningful), because only the integration tests exercise the code up to `req.SetXForwarded()`, which will overwrite some of the headers set by `addHeaders()`. I will open an issue to explain better this problem.


go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/go-kit/kit v0.13.0
 	github.com/go-kit/log v0.2.1
 	github.com/gobwas/glob v0.2.3
+	github.com/google/go-cmp v0.7.0
```
marco-m, line 10: In my opinion this is the standard to compare structs and maps, and should replace `github.com/pascaldekloe/goe/verify` (maybe for another PR :-)

proxy/http_integration_test.go
```diff
@@ -56,50 +57,73 @@ const (
...
+
+	clientHeader := http.Header{
+		legitHeader1: {"asdf"},
+		legitHeader2: {"qwerty"},
+		"Connection": {fmt.Sprintf("keep-alive, x-forwarded-for, x-forwarded-host, %s, %s, x-request-id, x-client-ip",
+			strings.ToLower(legitHeader1), strings.ToLower(legitHeader2))},
+		"X-Forwarded-For": {"3.3.3.3"},
+	}
+
+	t.Run("TrustedDownstream", func(t *testing.T) {
```
marco-m, line 107: You might not be used to this way of structuring tests. This is not the classic table-driven test, on purpose.

I learned this approach from the author of VictoriaMetrics: https://itnext.io/f-tests-as-a-replacement-for-table-driven-tests-in-go-8814a8b19e9e, which explains clearly the advantages: clicking on the line number of the failing test brings exactly ... to the failing test! Instead of bringing to the failing assertion and leaving the user to find out which test case in the table is failing :-)

If you disagree, I can go back to table-driven.

go.mod
tristanmorgan: `github.com/pascaldekloe/goe` does seem to be aging out.

marco-m: I forgot to add a comparison of the output. Consider:

```go
package mycmp

import (
	"testing"

	"github.com/google/go-cmp/cmp"
	"github.com/pascaldekloe/goe/verify"
)

type Foo struct {
	One   int
	Two   string
	Three []string
}

var (
	got = Foo{
		One:   1,
		Two:   "two",
		Three: []string{"1", "2", "3"},
	}

	want = Foo{
		One:   42,
		Two:   "fortytwo",
		Three: []string{"5", "4", "3", "2", "1"},
	}
)

func TestCmp(t *testing.T) {
	if diff := cmp.Diff(want, got); diff != "" {
		t.Errorf("%s\n--- want\n+++ got\n%s", "foos", diff)
	}
}

func TestVerify(t *testing.T) {
	verify.Values(t, "foos", got, want)
}
```

Output for cmp:
```
$ go test . -run TestCmp
--- FAIL: TestCmp (0.00s)
    cmp_test.go:32: foos
        --- want
        +++ got
          mycmp.Foo{
        - 	One: 42,
        + 	One: 1,
        - 	Two: "fortytwo",
        + 	Two: "two",
          	Three: []string{
        - 		"5",
        + 		"1",
        - 		"4",
        + 		"2",
          		"3",
        - 		"2",
        - 		"1",
          	},
          }
FAIL
```

Output for verify:
```
$ go test . -run TestVerify
--- FAIL: TestVerify (0.00s)
    cmp_test.go:37: verification for foos:
        /One: Got 1 (0x1), want 42 (0x2a)
        /Two: Got "two", want "fortytwo"
        /Three: Got 3 elements, want 5
FAIL
```

tristanmorgan (approved):
Thank you for the contribution.

